PR #5, opened 2024-02-13
Rely on upstream fixes to disable systematic desugaring of array assi…

msprotz:
…gnments, which in turn leads to much prettier code-gen.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,6 +30,7 @@ charon-test-%: $(CHARON_HOME)/tests/llbc/%.llbc | out/test-% all
 
 # Tests checked into the current repository
 .PHONY: phony
+.PRECIOUS: test/%/out.llbc
 test/%/out.llbc: phony
 	cd test/$* && $(CHARON) --errors-as-warnings && mv $*.llbc out.llbc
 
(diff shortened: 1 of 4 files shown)


msprotz:
@W95Psp is this using the latest krml master? it looks like this build hasn't picked up my upstream fixes... thanks!!

W95Psp:
hi @msprotz, indeed, dependencies are all pinned and locked!

If you have Nix installed, you can `nix flake update` to update every dependencies, and then commit the `flake.lock` file.
(or `nix flake lock --update-input INPUT` to update an input `INPUT`, with `INPUT` being a key of `inputs` in `flake.nix`)

msprotz:
Thanks! That already allowed me to uncovered another issue.

Is there a way to always pick the latest version available? Pinning a version of Charon has little value for me. For me, the main appeal of CI is being notified as soon as Charon changes something upstream so that I can catch up.

W95Psp:
The nightly job is picking the latest version of everything actually!
so if anything goes wrong with new Charon/Krml/F*/whatever, the nighlty job will fail

This PR/push job is testing against pinned dependencies, we can change that, but it is a bit weird with respect to the lockfile

msprotz:
Ah ok perfect! I didn't realize that was the purpose of the nightly. Sounds great then. Cheers!

PR #7, opened 2024-02-14
Several improvements in support of Kyber.

msprotz:
- Rather than bailing when trying to emit external, const-generic functions, see first if we can anticipate on C's array decay rules to replace e.g. t[N] in the signature with t*, in the hope that doing so removes all occurrences of const-generic variables. If it does, then we can emit this prototype! (And then length of the arrays will be known since the actual value of the const generic argument is passed at runtime.)
- For functions that do not pass this test, rely on an upstream fix to prop up the visibility of all the monomorphized type declarations that may be passed to the external at every call-site. This means that the external can now be implemented using a macro with the right type definitions in scope, which is essential for externals.
- Extend the syntax of configuration files with two new keys: include_in_h, a list of extra headers to be included in the .h for this file; and include_in_c, which is pretty self-explanatory.

This all allows removing a gigantic, sed-based hack for Kyber!

### DIFF
--- a/bin/main.ml
+++ b/bin/main.ml
@@ -127,8 +127,10 @@ Supported options:|}
   let files = Krml.Simplify.fixup_hoist#visit_files () files in
   let files = Krml.Simplify.misc_cosmetic#visit_files () files in
   let files = Krml.Simplify.let_to_sequence#visit_files () files in
+  Eurydice.Logging.log "Phase2.5" "%a" pfiles files;
   let files = Krml.Inlining.cross_call_analysis files in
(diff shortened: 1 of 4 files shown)


msprotz:
I can't believe this. I apparently managed to update the Nix flake. While drinking from my Nix mug. @pnmadelaine @W95Psp @twal

W95Psp:
Ahahaha, that is fantastic!! 😆🎉✨🎈🎆🎊

pnmadelaine:
:snowflake:

PR #11, opened 2024-05-15
Support for trait clause bounds

msprotz:
~This isn't quite working yet but I want to get this in place to start working through the code with @W95Psp.~

This is now on-par with what used to work before.

Let's try to block some time to go over this once you've had a chance to browse the list of changes. I've tried to add comments whenever I could, so hopefully some of this makes sense? Thanks!!


lib/AstOfLlbc.ml
```diff
@@ -172,6 +213,25 @@ module RustNames = struct
   let from =
     parse_pattern "core::convert::From<@T, @U>::from"
 
+  let into_u16 =
```
msprotz, line 216: @W95Psp I heard from @sonmarcho that there is a better, more concise way to do this, so this would be another interesting mini-task to add onto this PR -- refactor this bit to have more effective matching logic here, and in the main expression translation function

sonmarcho: What I do is actually quite simple, see [this](https://github.com/AeneasVerif/aeneas/blob/dc8ceedffa0945b84f81e318fa78f1e55b1731d2/compiler/ExtractBuiltin.ml#L13) and [this](https://github.com/AeneasVerif/aeneas/blob/dc8ceedffa0945b84f81e318fa78f1e55b1731d2/compiler/ExtractBuiltin.ml#L288)

lib/AstOfLlbc.ml
```diff
@@ -348,26 +409,75 @@ let mk_deep_copy (e: K.expr) (l: K.expr) =
...
   List.fold_left push_binder env ts
 
+(* Clause binders, which only appear as function parameters, and hold an unknown
+   trait method (dictionary-style). *)
+
+type clause_binder = {
+  clause_id: C.trait_clause_id;
+  item_name: string;
+  pretty_name: string;
+  sig_info: sig_info;
```
sonmarcho, line 460: What is the `sig_info` used for?

lib/AstOfLlbc.ml
```diff
@@ -348,26 +409,75 @@ let mk_deep_copy (e: K.expr) (l: K.expr) =
...
+  t: K.typ;
+}
+
+let push_clause_binder env { clause_id; item_name; t; sig_info; _ } =
+  { env with
+    (* Perhaps (?) important for diff calculations *)
+    cg_binders = (C.ConstGenericVarId.of_int max_int, t) :: env.cg_binders;
```
sonmarcho, line 467: I don't understand this

lib/AstOfLlbc.ml
```diff
@@ -348,26 +409,75 @@ let mk_deep_copy (e: K.expr) (l: K.expr) =
...
 
 let push_binders env (ts: C.var list) =
   List.fold_left push_binder env ts
 
+(* Clause binders, which only appear as function parameters, and hold an unknown
+   trait method (dictionary-style). *)
+
+type clause_binder = {
+  clause_id: C.trait_clause_id;
+  item_name: string;
```
sonmarcho, line 458: For every clause (e.g., `T : Clone`) you introduce one binder per item (e.g., the `clone` method coming from the `Clone` trait). Correct?

lib/AstOfLlbc.ml
```diff
@@ -566,47 +680,190 @@ let maybe_addrof (env: env) (ty: C.ty) (e: K.expr) =
   | _ ->
       K.(with_type (TBuf (e.typ, false)) (EAddrOf e))
 
+
+(** Handling trait clauses as dictionaries *)
+
+(* There are two ways that we skip synthesis of trait methods in function calls. The first one is if
+  a trait declaration is blocklisted. This happens if the trait has built-in support (e.g.
+  FnMut), or if the trait relies on unsupported features. The second way we skip trait methods
```
sonmarcho, line 688: "relies on unsupported features": do you have an example of trait method that you skip this way?

lib/AstOfLlbc.ml
```diff
@@ -566,47 +680,190 @@ let maybe_addrof (env: env) (ty: C.ty) (e: K.expr) =
...
+  "core::fmt::Debug";
+]
+
+(* Using tests/where_clauses_simple as an example.
+
+ fn double<T: Ops + Copy, U: Ops+Copy> (...)
+
+ this gets desugared to fn double<T,U> where
+   T: Ops,      <-- ClauseId 0 (required_methods: add, of_u32)
+   T: Copy,     <-- ClauseId 1 (builtin, so neither required nor provided methods)
```
sonmarcho, line 715: Side note: the important point with `Copy` is that it is a supertrait of `Clone`, so if you implement `Copy` you implement `Clone` (i.e., you have a `clone` method). We may implement a micro-pass in Charon which removes instances of `Copy` whenever possible, and adds a builtin for instances of `Clone` which actually come from `Copy`, so that you don't need to implement any custom support here.

lib/AstOfLlbc.ml
```diff
@@ -566,47 +680,190 @@ let maybe_addrof (env: env) (ty: C.ty) (e: K.expr) =
...
+        (C.TraitClauseId.to_int clause_id)
+        (String.concat " ++ " (List.map C.show_ty (clause_generics.C.types)))
+        (List.length trait_decl.C.required_methods)
+        (List.length trait_decl.C.provided_methods);
+
+      (* The type variable from the function that the trait clause is applied to *)
+      let tvarid =
+        match clause_generics.C.types with
+        | [ TVar tvarid ] -> tvarid
+        | _ -> failwith "Clause above does not refer to a type var? confused"
```
sonmarcho, line 744: This message should be more general: there can be strictly more than one type variable

lib/AstOfLlbc.ml
```diff
@@ -566,47 +680,190 @@ let maybe_addrof (env: env) (ty: C.ty) (e: K.expr) =
...
+            (clause_id, item_name), (tvarid, trait_decl.C.name, decl.C.signature)
+        | None ->
+            failwith ("TODO: handle provided trait methods, like " ^ item_name)
+            ) trait_decl.C.provided_methods
+    end
+  ) trait_clauses
+
+(* Interpret a Rust function type, with trait bounds, into the krml Ast, providing:
+   - the number of polymorphic type variables
+   - the cg types, which only contains the original Rust const generic variables (not trait methods)
```
sonmarcho, line 764: What do you mean by "not trait methods"?

lib/AstOfLlbc.ml
```diff
@@ -566,47 +680,190 @@ let maybe_addrof (env: env) (ty: C.ty) (e: K.expr) =
...
+    let sig_info = {
+      n_type_args = List.length signature.generics.types - 1;
+      n_cg_args = List.length signature.generics.const_generics;
+    } in
+    { pretty_name; t; clause_id; item_name; sig_info }
+  ) clause_mapping
+
+
+(* Transforms a lookup result into a usable type, taking into account the fact that the internal Ast
+   is ML-style and does not have zero-argument functions. *)
```
sonmarcho, line 813: What do you do for the functions defining the bodies of globals (which are functions with 0 arguments)?

lib/AstOfLlbc.ml
```diff
@@ -628,39 +896,77 @@ let fn_ptr_is_opaque env (fn_ptr: C.fn_ptr) =
...
+  L.log "Calls" "%sis_array_map: %b" depth (RustNames.is_array_map env fn_ptr);
+  L.log "Calls" "%s--> %d type_args, %d const_generics, %d trait_refs" depth
     (List.length type_args) (List.length const_generic_args) (List.length trait_refs);
 
   let type_args, const_generic_args, trait_refs =
     match func with
-    | TraitMethod ({ generics = { types; const_generics; trait_refs; _ }; _ }, _, _) ->
-        types @ type_args, const_generics @ const_generic_args, trait_refs @ trait_refs
+    | TraitMethod ({ generics = { types; const_generics; trait_refs = trait_refs'; _ }; _ }, _, _) ->
+        (* TODO: why is the order fumbled here? *)
```
sonmarcho, line 918: What do you mean?

lib/AstOfLlbc.ml
```diff
@@ -868,16 +1181,30 @@ let rec expression_of_raw_statement (env: env) (ret_var: C.var_id) (s: C.raw_sta
       Krml.Helpers.with_unit K.(EAssign (dest, maybe_addrof env ty (with_type t (EBufRead (e1, e2)))))
 
   | Call { func = FnOpRegular fn_ptr; args; dest; _ }
-    when false && Charon.NameMatcher.match_fn_ptr env.name_ctx RustNames.config RustNames.from fn_ptr ->
+    when (
+      Charon.NameMatcher.match_fn_ptr env.name_ctx RustNames.config RustNames.from_u16 fn_ptr ||
```
sonmarcho, line 1185: You can put those in a map:
https://github.com/AeneasVerif/charon/blob/ed5047fd07e6c6aa8e7b9f352637f91977adacbe/charon-ml/src/NameMatcher.ml#L1186

lib/AstOfLlbc.ml
```diff
@@ -868,16 +1181,30 @@ let rec expression_of_raw_statement (env: env) (ret_var: C.var_id) (s: C.raw_sta
...
       (* Special treatment: From<T, U> becomes a cast. *)
       let matches p = Charon.NameMatcher.match_fn_ptr env.name_ctx RustNames.config p fn_ptr in
       let w: Krml.Constant.width =
-        if matches RustNames.from_u16 then UInt16
-        else if matches RustNames.from_u32 then UInt32
-        else if matches RustNames.from_u64 then UInt64
-        else if matches RustNames.from_i16 then Int16
-        else if matches RustNames.from_i32 then Int32
-        else if matches RustNames.from_i64 then Int64
+        if      matches RustNames.from_u16 || matches RustNames.into_u16 then UInt16
```
sonmarcho, line 1202: Same remark as above

test/where_clauses_simple/src/main.rs
```diff
@@ -0,0 +1,26 @@
+trait Ops {
```
sonmarcho, line 1: An interesting example (which I don't think you handle yet) is supertraits (or "traits with parents" to reuse Charon's terminology).
Ex.:
```rust
trait T1 { ... }
trait T2 : T1 { ...}
```

Those happen in many places in Rust with `Copy`/`Clone`, `Eq`/`PartialEq`, `Ord`/`PartialOrd`, etc.

lib/AstOfLlbc.ml
```diff
@@ -628,39 +896,77 @@ let fn_ptr_is_opaque env (fn_ptr: C.fn_ptr) =
...
+        let name = string_of_name env (env.get_nth_trait_decl trait_ref.trait_decl_ref.trait_decl_id).name in
+
+        if List.mem name blocklisted_trait_decls then
+          []
+
+        else
+          let trait_impl: C.trait_impl =
+            match trait_ref.trait_id with
+            | TraitImpl impl_id -> env.get_nth_trait_impl impl_id
+            | _ -> failwith ("impossible: " ^ C.show_trait_ref trait_ref)
```
W95Psp, line 956: Why is this supposed to be impossible?

The following triggers a case where the `trait_impl` is a `Clause`:
```rust
trait Marker {}

trait Trait {
    fn f();
}

impl<T: Marker> Trait for T {
    fn f() { }
}

fn bug<T: Marker>(x: T) {
    T::f()
}
```

Here, in function `bug`, the call `T::f` is actually `(impl_0::<Clause 0>)::f`, where `Clause 0` refers to the generic bound `<T: Marker>` on function `bug`.
The concrete impl block `impl<T: Marker> Trait for T` has one generic `T` which requires an impl for `Marker`.

If I understand correctly, we are here trying to construct a `trait_impl` for every bound that occurs both on the function itself but also on the impl block?

Looking at the type `C.trait_impl`, I wonder if that type can represent such local bounds (here we have nothing but a bound variable).

W95Psp: Actually, this example also triggers the same "impossible" error:
```rust
trait Trait {}

fn f<T: Trait>() {}
fn g<T: Trait>() {
    f::<T>()
}
```

lib/AstOfLlbc.ml
msprotz: Is there a way to recover the values for which @T and @U matched? I guess that's what I had in mind when I wrote "a better way"

lib/AstOfLlbc.ml
```diff
@@ -26,21 +26,59 @@ module L = Logging
...
 
   (* Needed by the name matching logic *)
   name_ctx: Charon.NameMatcher.ctx;
   generic_params: C.generic_params;
 
-  (* To compute DeBruijn indices *)
+  (* We have three binding scopes, all in DeBruijn indices.
+     - const-generic ("cg") binders
+     - type binders
+     - expression binders, which typically starts with a repeat of the cg
```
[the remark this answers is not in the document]
msprotz: Added a comment to clarify.

lib/AstOfLlbc.ml
msprotz: The sig_info was to provide a type signature for the function that was found as a result of the lookup. It is now replaced by `ts` (a type scheme, which counts the number of type variables and cg variables at the beginning of this signature), and `t`, a type that goes under this type scheme.

lib/AstOfLlbc.ml
msprotz: This is gone. It looks like I didn't understand this either.

lib/AstOfLlbc.ml
msprotz: Correct.

lib/AstOfLlbc.ml
msprotz: added an example of a trait method that I skip

lib/AstOfLlbc.ml
msprotz: that sounds nice!

lib/AstOfLlbc.ml
msprotz: yup this is now gone

lib/AstOfLlbc.ml
msprotz: Removed this. It's clear from the following lines that binders for the dictionary-style passing of trait methods are in arg_types.

lib/AstOfLlbc.ml
msprotz: Dedicated treatment (see https://github.com/AeneasVerif/eurydice/blob/main/lib/AstOfLlbc.ml#L1152-L1166) -- I directly call into expression_of_raw_statement without going through lookup_function

lib/AstOfLlbc.ml
msprotz: What I meant is that trait_refs' was concatenated in the opposite order as the concatenation of `const_generics` -- this has since then been fixed.

test/where_clauses_simple/src/main.rs
msprotz: Yes these are in the (disabled) where_clauses trait.

lib/AstOfLlbc.ml
msprotz: This has since then been fixed with support for trait bounds implemented as dictionary-passing style. Thanks for the minimal repro.

lib/AstOfLlbc.ml
msprotz: Created https://github.com/AeneasVerif/eurydice/issues/20 as a followup

msprotz:
I'm merging this and opened #20 to track the specific issue of improving the name matcher logic.

PR #21, opened 2024-06-04
nix: add a dev shell

Nadrieril:
This PR does two things:
- Adds a dev shell for nix users like me so we can easily develop on eurydice;
- ~Makes it so the nix builder (and thus the CI) reports warning as errors, because CI was missing errors like non-exhaustive matches. I set all warnings to be errors which might be a bit strong; we can tweak as needed.~

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -47,6 +47,8 @@
 
           src = ./.;
 
+          OCAMLPARAM = "_,warn-error=+A"; # Turn all warnings into errors.
+
           nativeBuildInputs = [gnugrep];
 
           propagatedBuildInputs = [krml charon-ml ocamlPackages.terminal ocamlPackages.yaml];
--- a/flake.nix
+++ b/flake.nix
@@ -81,5 +81,16 @@
         version = self.rev or "dirty";
       };
       checks.default = packages.default.tests;
+      devShells.default = pkgs.mkShell {
+        packages = [
+          pkgs.ocamlPackages.ocaml
+          pkgs.ocamlPackages.ocamlformat
+          pkgs.ocamlPackages.menhir
+        ];
+
+        inputsFrom = [
+          self.packages.${system}.default
+        ];
+      };
     });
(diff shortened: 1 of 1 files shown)


Nadrieril:
With this Ci change we should likely coordinate breaking changes in charon with eurydice like we do for aeneas. The idea would be that if charon makes a breaking change, we wait for the appropriate PR to be ready on the eurydice side before merging both. Then for completeness we should also add `kyber` to eurydice's CI.

Nadrieril:
I'm glad you know something about ocaml flags, I was just copying what we have for aeneas. I'll try some things tomorrow

Nadrieril:
I removed the part about build flags, I'll figure that out later. I'd like the dev shell to be merged first because it makes my life easier

msprotz:
Looking forward to the subsequent PR for warnings, and happy to chat about it with tomorrow morning if that helps (just ping me)

Nadrieril:
It seems I failed to remove the part about build flags x) I might have gotten confused between this repo and my fork, oops.

msprotz:
Yeah I noticed it's fine. Can you submit a followup when you get a chance so that we can fix the build flag story?

Nadrieril:
I wasn't particularly excited to dig into ocaml docs. Do you happen to know the right incantation? Afaik the only reason warnings are silenced is because we cuild in release mode, so maybe we only need to change the `dune` file?

msprotz:
I vote for this in the various `dune` files:

```
(env
  (_
    (flags (:standard -warn-error -A -w @1-2@3-7@8..12@14..21@23..29-30@31..38-39-40-41@43@57))))
```

after https://github.com/FStarLang/karamel/blob/master/lib/dune#L27 -- the `_` should ensure this is picked for both dev and release

msprotz:
although -warn-error -A is probably redundant here -- I would only keep the `-w` part

PR #25, opened 2024-06-11
Preliminary support for parent clauses

msprotz:
This should be enough to unblock https://github.com/cryspen/libcrux/pull/308

This is not enough to make test-where_clauses go through, so I still want to leave this open until I get my testcase to work

### DIFF
--- a/lib/AstOfLlbc.ml
+++ b/lib/AstOfLlbc.ml
@@ -32,7 +32,7 @@ open Krml.PrintAst.Ops
    in this order, the first two being only ever inserted upon entering a function
    definition. *)
 type var_id =
-  | TraitClauseMethod of C.trait_clause_id * string * K.type_scheme
+  | TraitClauseMethod of C.trait_instance_id * string * K.type_scheme
   | ConstGenericVar of C.const_generic_var_id
   | Var of C.var_id * C.ety (* the ety aids code-generation, sometimes *)
 
@@ -47,10 +47,6 @@ type env = {
   (* Needed by the name matching logic *)
   name_ctx: Charon.NameMatcher.ctx;
   generic_params: C.generic_params;
-
-  (* When in a function such as fn f<T: Foo<42>>, we need to remember that Foo is instantiated over
-     42 so as to pass the correct arguments. *)
-  clause_arguments: (C.TraitClauseId.id * C.const_generic list) list;
 
   (* We have three lists of binders, which allow us to go from a Rust variable
      to a corresponding krml AST variable; everything is in De Bruijn, so
@@ -478,7 +474,7 @@ let push_binders env (ts: C.var list) =
    trait method (dictionary-style). *)
 
 type clause_binder = {
-  clause_id: C.trait_clause_id;
+  clause_id: C.trait_instance_id;
   item_name: string;
   pretty_name: string;
   ts: K.type_scheme;
@@ -755,7 +751,7 @@ let blocklisted_trait_decls = [
  the types we obtain by looking up the trait declaration have Self as 0
  (DeBruijn).
  *)
-let build_trait_clause_mapping env (trait_clauses: C.trait_clause list) =
+let rec build_trait_clause_mapping env (trait_clauses: C.trait_clause list) =
   List.concat_map (fun tc ->
     let { C.clause_id; trait_id; clause_generics; _ } = tc in
     let trait_decl = env.get_nth_trait_decl trait_id in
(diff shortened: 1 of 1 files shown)


lib/AstOfLlbc.ml
```diff
@@ -782,17 +778,26 @@ let build_trait_clause_mapping env (trait_clauses: C.trait_clause list) =
...
         | None ->
             failwith ("TODO: handle provided trait methods, like " ^ item_name)
-            ) trait_decl.C.provided_methods
+      ) trait_decl.C.provided_methods @
+      List.flatten (List.mapi (fun i (parent_clause: C.trait_clause) ->
+        (* Mapping of the methods of the parent clause *)
+        let m = build_trait_clause_mapping env [ parent_clause ] in
+        List.map (fun ((clause_id, m), v) ->
+          (* This is the parent clause `i` of `clause_id` -- see comments in charon/types.rs  *)
+          let id = C.(ParentClause (clause_id, trait_id, TraitClauseId.of_int i)) in
```
msprotz, line 797: slightly strange that I have to rebuild this myself, please check that I'm doing the right thing

lib/AstOfLlbc.ml
```diff
@@ -988,16 +996,9 @@ let rec expression_of_fn_ptr env depth (fn_ptr: C.fn_ptr) =
...
         L.log "Calls" "%s--> this is a trait method" depth;
-        let clause_const_generics =
-          match trait_id with
-          | Clause id when false ->
-              List.assoc id env.clause_arguments
-          | _ ->
-              []
-        in
-        types @ type_args, clause_const_generics @ const_generics @ const_generic_args, trait_refs' @ trait_refs
+        types @ type_args, const_generics @ const_generic_args, trait_refs' @ trait_refs
```
msprotz, line 1001: this is an unrelated change that removes dead code

lib/AstOfLlbc.ml
```diff
@@ -1025,57 +1026,68 @@ let rec expression_of_fn_ptr env depth (fn_ptr: C.fn_ptr) =
          anyhow. *)
       []
     else
-      List.concat_map (fun (trait_ref: C.trait_ref) ->
-        let name = string_of_name env (env.get_nth_trait_decl trait_ref.trait_decl_ref.trait_decl_id).name in
+      (* MUST have the same structure as build_trait_clause_mapping *)
```
msprotz, line 1029: this populates the trait method references at call-site, so we need to materialize function pointers exactly as expected by the signature, which is determined by `build_trait_clause_mapping`

lib/AstOfLlbc.ml
```diff
@@ -1025,57 +1026,68 @@ let rec expression_of_fn_ptr env depth (fn_ptr: C.fn_ptr) =
...
+              (* This must be in agreement, and in the same order as build_trait_clause_mapping *)
+              List.map (fun (item_name, decl_id) ->
+                let fn_ptr: C.fn_ptr = {
+                  func = TraitMethod (trait_ref, item_name, decl_id);
+                  generics = Charon.TypesUtils.empty_generic_args
+                } in
+                let fn_ptr = fst3 (expression_of_fn_ptr env (depth ^ "  ") fn_ptr) in
+                fn_ptr
+              ) (trait_impl.required_methods @ trait_impl.provided_methods) @
+              build_trait_ref_mapping ("  " ^ depth) trait_ref.generics.trait_refs
```
msprotz, line 1053: here, the idea is that after filling out required and provided methods, this should match what is done in build_trait_clause_mapping for the parent clauses -- but I'm not sure this is the right way, so please check!

lib/AstOfLlbc.ml
```diff
@@ -1025,57 +1026,68 @@ let rec expression_of_fn_ptr env depth (fn_ptr: C.fn_ptr) =
...
+
+          | Clause _ as clause_id ->
+              (* Caller it itself polymorphic and refers to one of its own clauses to synthesize
+                 the clause arguments at call-site. We must pass whatever is relevant for this
+                 clause, *transitively* (this means all the reachable parents). *)
+              let rec relevant = function
```
msprotz, line 1059: insofar as I understand, when the caller is clause-polymorphic, and the callee is clause-polymorphic too, then I only see a single `Clause _` reference at call-site, and it's up to me to remember that this trait bound requires parent bounds too, and materialize method references for the parent, too

lib/AstOfLlbc.ml
```diff
@@ -1582,13 +1594,6 @@ let decls_of_declarations (env: env) (d: C.declaration_group): K.decl list =
                   let clause_mapping = build_trait_clause_mapping env signature.C.generics.trait_clauses in
                   debug_trait_clause_mapping env clause_mapping;
 
-                  (* In definition fn f<T: Foo<42>>, associates [ 42 ] to clause Foo *)
```
msprotz, line 1585: just removing dead code

msprotz:
The next phase requires a little more work: I need to be handle trait clause bounds in trait impl methods, which is non-trivial (see test/where_clauses), so I'm going to leave it at this for now -- I cannot handle trait instances of the form Parent(Self, ...) yet.

So let's merge this since it already improves the state of things, once I get a green-light that what I'm doing vaguely makes sense.

lib/AstOfLlbc.ml
Nadrieril: Mm yes, charon models a clause as "containing" its parent clauses, so we don't mention the parent clauses at call site. I think you should never think of a clause without its parents, so there should be no need for you to explore the parents. What are you trying to do here exactly?

lib/AstOfLlbc.ml
Nadrieril: aybe you could use something more structured than a list then, e.g. a tree that follows the parent relationships

lib/AstOfLlbc.ml
Nadrieril: So the assumption is that the list of trait_refs clauses matches the required list of clause arguments to the trait? That feels correct. I've been wanting to provide a better API for things like that.

lib/AstOfLlbc.ml
msprotz: well I'm recursing into the parent clause and it has its own clause id relative to itself, without the ParentClause around it

lib/AstOfLlbc.ml
msprotz: ok, well I believe this should provide a good use-case for that :-)

lib/AstOfLlbc.ml
msprotz: I think it's fine, yes

msprotz:
Thanks for the comments @Nadrieril it's good to know my assumptions are correct. I'll merge this for now, and ping me if you want to discuss this ParentClause business a little bit more (but the more I think about it, the more I think there's nothing wrong on the Charon side)

PR #29, opened 2024-06-24
Setup formatting

Nadrieril:
Wdyt?


Makefile
```diff
@@ -36,6 +36,11 @@ test-%: test/%/out.llbc out/test-%/main.c | all
 out/%:
 	mkdir -p $@
 
+.PHONY: format
```
msprotz, line 39: can you just integrate this into `format-apply`, and try to notice missing formatting in format-check? that way it's unified for C and OCaml.

msprotz:
looks good to me, but let's integrate this with the existing formatting targets for C before merging. thanks!!

Makefile
Nadrieril: Oh yep, I hadn't seen them

Nadrieril:
Like this?

msprotz (approved):
Looks good to me thank you.

The last remark that I have re. formatting is, if there's a way to *always* break to a new line for if-then-else (and match branches) I'd love it, otherwise, let's merge and move on. Like

```
if e then
  e1
else
  e2
```

even if it would fit on one line.

Thanks for your hard work on this! I know these topics are frustrating so I'm glad to see this converging.

Nadrieril:
I found an option that does what you want for if-then-else. Wdyt?

PR #34, opened 2024-06-28
Don't silence errors in CI

Nadrieril:
As discussed [here](https://github.com/AeneasVerif/eurydice/pull/21#issuecomment-2183062622). I tested locally and CI indeed fails on e.g. a non-exhaustive pattern match.


msprotz:
Great. Thank you!

PR #41, opened 2024-07-11
Propagate target attribute

karthikbhargavan:
This PR propagates target attributes to the C code.
In particular, if you add to the config.yaml file:
```
- name: libcrux_sha3_avx2
   target: "avx2"
   ...
```

Then the file `libcrux_sha3_avx2.c` will contain:
```
KRML_ATTRIBUTE_TARGET("avx2")
```
before every function in the file.

This has only been tested on libcrux, so it may need a local test file and documentation.


bin/main.ml
```diff
@@ -224,6 +224,25 @@ Supported options:|}
...
+      match config with
+      | None -> ""
+      | Some c -> 
+	match List.find_opt (fun (x:file) -> x.name = f) c with
+	| None -> ""
+	| Some f -> f.target in
+    f, List.filter_map (fun d ->
+      match d with
+      | Krml.Ast.DFunction (cc,fl,x,y,t,l,b,e) when target_attribute <> "" ->
+          Some (Krml.Ast.DFunction (cc,fl@[Target(target_attribute)],x,y,t,l,b,e))
```
msprotz, line 239: you can just write `Target target_attribute` here

msprotz (approved):
looks good to me, thanks!

please add a comment above that new rewriting phase saying what it does, and then you're good to go

PR #60, opened 2024-08-21
Check kyber in CI

Nadrieril:
Since I do it in the charon side, it makes sense to do it here as well.


msprotz (approved):

msprotz:
why is this test failing? is it known?

Nadrieril:
Yeah it seems I need to set the right version of karamel like I did for charon. This is a known limitation of nix flakes

msprotz:
ok then if you think this is good please go ahead and merge

Nadrieril:
(this is blocked on the infinite loop that was caused by the latest charon PR)

Nadrieril:
This is too slow to be practical. Let's move to our self-hosted runner, it has good nix caching

PR #61, opened 2024-08-27
Add another core_array_equality to eurydice_glue.h

franziskuskiefer:
Add another `core_array_equality` function.


msprotz:
Cool, confirm that a1 and a2 are arrays, not slices? From the name it looks like the arguments may be slices.

franziskuskiefer:
Ha good catch. It's indeed a slice. I'll update later.

msprotz (commented):
That's better, but `t*` is not going to work if `t` is a non-scalar type, e.g. the "pointerized" version of `int[2]` is `(int*)[2]`, not `int[2]*`.

Do you need that cast here? If so, we need custom support for this in code-gen (which I'm happy to implement).

msprotz (approved):

PR #63, opened 2024-08-27
Prettify assertions

msprotz:


msprotz:
will fix later -- this is just to save work in progress

PR #72, opened 2024-09-02
some wrapping operations over i16s

karthikbhargavan:


msprotz (changes requested):
This doesn't work because x + y is undefined behavior in case there is an overflow in C (and does not wrap -- it just UBs). See https://wiki.sei.cmu.edu/confluence/display/c/INT32-C.+Ensure+that+operations+on+signed+integers+do+not+result+in+overflow -- I'm not sure what the right way to implement this is.

(The original comment mean "Unsigned integers have wraparound overflow semantics in C so no special treatment is needed.")

PR #76, opened 2024-09-24
Carry comments from Rust to C

msprotz:


msprotz:
To be re-run once fstarlang/karamel#483 gets merged

msprotz:
@Nadrieril should we make kyber a requirement before merging a PR here? Or is it not made a required check to allow asynchronous fixing of the two?

msprotz:
Just realized this got merged without kyber coming back green

Nadrieril:
It's not a requirement because libcrux doesn't have processes in place to make it reliably be green.

PR #83, opened 2024-11-13
Update charon

Nadrieril:
Companion PR to https://github.com/AeneasVerif/charon/pull/457 which changes how we represent places in charon. In particular, charon now stores the type of each place. I use this to simplify `expression_of_place`.

### DIFF
--- a/lib/AstOfLlbc.ml
+++ b/lib/AstOfLlbc.ml
@@ -521,15 +521,15 @@ let expression_of_const_generic env cg =
   | C.CgVar id -> expression_of_cg_var_id env id
   | C.CgValue l -> expression_of_literal env l
 
-let expression_of_place (env : env) (p : C.place) : K.expr * C.ety =
-  let { C.var_id; projection } = p in
-  let e, ty = expression_and_original_type_of_var_id env var_id in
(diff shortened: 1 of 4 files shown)


lib/AstOfLlbc.ml
```diff
@@ -521,95 +517,91 @@ let expression_of_const_generic env cg =
...
-                  assert (List.length ts = n);
-                  ts, List.nth ts i
-              | _ ->
-                  assert (List.length tys = 1);
-                  L.log "AstOfLlbc" "typ is: %a" ptyp e.typ;
-                  failwith "impossible: mismatch ProjTuple/TTuple"
+            let ts =
+              match sub_e.typ with
+              | TTuple ts -> ts
+              | _ -> assert false
```
msprotz, line 599: so the issue with tuples of length 1 has been fixed upstream?

msprotz: ah I see, the early-return case is still up there

msprotz (approved):
Great. This new representation is much better and lends itself to recursion, which is hella more intuitive. Thanks for the patch, much appreciated! Have a great weekend.

PR #85, opened 2024-11-18
Deduplicate `fstar` input in flake

Nadrieril:
The flake pinned `fstar` in two places (one directly and once inside karamel), which means if they ever got out of sync we'd have to build it twice. This instead reuses the pinned version in karamel.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -107,46 +107,10 @@
         "type": "indirect"
       }
     },
-    "flake-utils_4": {
-      "inputs": {
-        "systems": "systems_4"
-      },
-      "locked": {
-        "lastModified": 1692799911,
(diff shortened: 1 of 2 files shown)


Nadrieril:
@msprotz are there cases where you would want to use a different version of fstar in eurydice than the one pinned by the karamel version you're using?

msprotz:
Eurydice should not need F*

msprotz:
Let me fix https://github.com/FStarLang/karamel/issues/496 real quick

Nadrieril:
Can we just merge this and you can remove the fstar input once it's not needed?

msprotz:
The alternative is to export a dummy FSTAR_HOME, by the way, but this is really silly

PR #92, opened 2024-11-21
Fix an overly specific pass

msprotz:
As explained in comments, slice_index does not need to be rewritten in case the type of indices is an array. A nano-pass was recovering from the over-zealous `pass_by_ref` phase from krml, but for some reason was crafted to only trigger on a very specific form of expressions. This fixes it.


msprotz:
This pass was trying to recover from some poor, earlier, code-gen decisions. Closing in favor of a proper fix.

PR #113, opened 2024-12-19
Improve on assert recognition

msprotz:
@sonmarcho FYI a new form of assert is now generated by Charon: `if not e1 then abort msg else ()` is now recognized as `Eurydice_assert(e1, msg)`.


sonmarcho:
Ok, I missed this one. I will also add it on Charon's side.

msprotz:
No need to change anything, I can now deal with it too and it will be emitted properly. I think the current situation is ok. Thanks!

PR #114, opened 2024-12-19
Honor user-provided enum numbering schemes

### LINKED ISSUE #102: Enums are not translated correctly
The following Rust code

```rust
enum Error1 {
    Reason1 = 1,
    Reason2 = 2,
}

enum Error2 {
    Reason1 = 3,
    Reason2 = 4,
}
```

Is translated into

```c
#define eurydice_tests_Reason1 0
#define eurydice_tests_Reason2 1

typedef uint8_t eurydice_tests_Error1;

typedef eurydice_tests_Error1 eurydice_tests_Error2;
```

msprotz:
This should fix #102 and needs the branch of the same name for krml.

I'll wait for circus CI to go back to green before I try this out -- want to make sure I don't break any big use-case.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -11,11 +11,11 @@
         "rust-overlay": "rust-overlay"
       },
       "locked": {
-        "lastModified": 1733929967,
-        "narHash": "sha256-RaCG23BtUsa6iCoFBgS5Uv7FdLzKQX8zPFHBDvl/v58=",
-        "owner": "aeneasverif",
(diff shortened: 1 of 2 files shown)


msprotz:
Now that circus is back to green (thanks @sonmarcho) I was able to fix a few more things, and now I observe no regressions. I'll merge this right after https://github.com/FStarLang/karamel/pull/507, its companion PR.

PR #120, opened 2024-12-30
Tweak: Use a helper function

Nadrieril:
Use a helper function instead of constructing a context by hand. This makes it so https://github.com/AeneasVerif/charon/pull/509 doesn't break Eurydice.


msprotz:
Neat! Thanks

PR #121, opened 2025-01-02
Update charon

Nadrieril:
Companion PR to https://github.com/AeneasVerif/charon/pull/512.

There were some formatting changes, I'm surprised. Is it that we're using different settings or that auto-formatting is no longer happening on your side?

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -11,11 +11,11 @@
         "rust-overlay": "rust-overlay"
       },
       "locked": {
-        "lastModified": 1734731548,
-        "narHash": "sha256-1dQ35aUZ2P6xIV4k8ueOcY+q7Et6puFccMhJ/0NJWXM=",
-        "owner": "AeneasVerif",
+        "lastModified": 1735820899,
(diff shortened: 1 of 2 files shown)


msprotz:
Weird. I formatted recently (786ce0d4777fe138608c7616ace81b9c45d67e44) so I'm surprised you had so much diff.

Nadrieril:
It could be because I use ocamlformat `0.26.0` while you use `0.26.2`. The version is set in the `.ocamlformat` file for a reason (I have to tweak it to get it to work). Could you try formatting yourself now? If this reverts my changes then it's the version difference.

Nadrieril:
`make format-apply` will use whichever version of `ocamlformat` is in `$PATH`. If like me you're using the nix-provided environment then this uses a known version (`0.26.0` in fact), but otherwise it uses the system-installed version, like we do for the rest of the ocaml toolchain.

Nadrieril:
`ocamformat` 0.26.2 is available in nix, so if that's the cause of the formatting differences then I can just set my environment to use it.

Issue #125, opened 2025-01-03
TraitTypes support


franziskuskiefer:
Using trait types doesn't work when the trait is used in a type bound or in any other way.

```rust
trait T {
    type C;
    fn new() -> Self::C;
}
struct S {}
impl T for S {
    type C = [u8; 4];

    fn new() -> Self::C {
        Self::C::default()
    }
}

fn fun<Type: T>() {}

fn caller() {
    fun::<S>();
}

fn associated_type() -> [u8; 4] {
     S::new()
}
```

[generates](https://github.com/AeneasVerif/eurydice/blob/8e112cd3065d2c1eb6c023cd37111300dbf9fc9a/lib/AstOfLlbc.ml#L394)

```
void eurydice_tests_caller(void)
{
  KRML_HOST_EPRINTF("KaRaMeL abort at %s:%d\n%s\n",
    __FILE__,
    __LINE__,
    "Eurydice error: Failure(\"TODO: TraitTypes Self::C\")\n");
  KRML_HOST_EXIT(255U);
}

void eurydice_tests_associated_type(uint8_t ret[4U])
{
  KRML_HOST_EPRINTF("KaRaMeL abort at %s:%d\n%s\n",
    __FILE__,
    __LINE__,
    "Eurydice error: Failure(\"TODO: TraitTypes eurydice_tests::{eurydice_tests::T for eurydice_tests::S}::C\")\n");
  KRML_HOST_EXIT(255U);
}
```

Nadrieril:
Will get fixed by https://github.com/AeneasVerif/charon/issues/127.

franziskuskiefer:
Great! Do you have an ETA for those changes?

Nadrieril:
This month if all goes well, the next if not.

msprotz:
@franziskuskiefer and I had a chat this morning, and considering that Franziskus has a workaround already (and contrary to what I said earlier), I am now dropping this and will let you fix it upstream -- it's only a matter of weeks, not worth the extra complexity

Nadrieril:
Fixed by https://github.com/AeneasVerif/eurydice/pull/153

PR #126, opened 2025-01-07
Preliminary support for associated types

msprotz:
This converts on the fly associated types to extra type parameters in function declarations, to be resolved later on as type parameters in scope.


Nadrieril:
Ah yes, this error is triggered if you attempt to substitute a `generic_params`, because there's a high chance the resulting value would be nonsensical: here it would be a signature that claims to bind some generic variables but you actually just substituted them away in e.g. the input/output types.

Nadrieril:
Actually I tried to make the error message better and ended up removing this restriction altogether since we generally want to be able to substitute things that contain `generic_params`. This should be fixed by https://github.com/AeneasVerif/charon/pull/518.

msprotz:
Wonderful, thank you. Closing per my comment in #125

PR #131, opened 2025-01-10
Commit test output

Nadrieril:
As discussed privately, this PR commits the `.c` and `.h` files generated for our tests. This will make it easier to spot behavior changes. CI will ensure the committed test output matches what's being generated.


Nadrieril:
@protz This fails with
```diff
-typedef int8_t core_cmp_Ordering;
+typedef uint8_t core_cmp_Ordering;
```
in some tests. Do you know why eurydice picks a different type here compared to my machine? Could this be an architecture difference? I am particularly surprised because `Ordering` has [a defined representation](https://doc.rust-lang.org/std/cmp/enum.Ordering.html) which includes `-1` so this should never be `uint`.

Nadrieril:
Oh that's because I was using an old commit.

flake.nix
```diff
@@ -73,7 +73,24 @@
...
                   export CHARON_HOME=./charon
 
+                  # Move the committed test outputs out of the way
+                  mv out out-comitted
+
+                  # Run the tests
                   make -o all test
+
+                  # Clean generated files that we don't want to compare.
+                  rm out/**/a.out
```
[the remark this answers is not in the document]
Nadrieril: It executes in bash, which may indeed not have that by default. I should shopt

Makefile
```diff
@@ -6,6 +6,10 @@ CHARON		?= $(CHARON_HOME)/bin/charon
 BROKEN_TESTS		= step_by where_clauses chunks mutable_slice_range closure issue_49 issue_37 issue_105
 TEST_DIRS		= $(filter-out $(BROKEN_TESTS),$(subst test/,,$(shell find test -maxdepth 1 -mindepth 1 -type d)))
 
+# Enable `foo/**` glob syntax
+SHELL := bash -O globstar
```
msprotz, line 10: @Nadrieril Son started reporting issues with this:

```
bash: line 0: globstar: invalid shell option name
bash: line 0: globstar: invalid shell option name
```

@sonmarcho is possibly on OSX where the default version of make is rather old -- we forbid this in some other projects https://github.com/hacl-star/hacl-star/blob/main/Makefile#L46-L49

Nadrieril: hm, that looks like a `bash` difference, not a `make` difference

Nadrieril: Indeed an OSX difference: https://apple.stackexchange.com/questions/291287/globstar-invalid-shell-option-name-on-macos-even-with-bash-4-x . I guess I should avoid the glob and do a `find | xargs` dance?

msprotz: sorry I meant bash -- this is the same issue, OSX ships a very outdated bash for GPLv3 reasons

this is the bash version that is bad in OSX:

jonathan@absinthe:~/Code/noise-star/src (main) $ /bin/bash --version
GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin24)

if you can detect that and error out with a suggestion to `brew install bash`, then I'm fine with it (I personally run with brew bash and everyone else should)

```
jonathan@absinthe:~/Code/noise-star/src (main) $ /bin/bash
bash: complete: nosort: invalid option name

The default interactive shell is now zsh.
To update your account to use zsh, please run `chsh -s /bin/zsh`.
For more details, please visit https://support.apple.com/kb/HT208050.
jonathan@absinthe:~/Code/noise-star/src (main) $ echo $BASH_VERSION
3.2.57(1)-release
```

Nadrieril: Do you know a reliable way to detect that?

msprotz: since Apple's bash is frozen in time I think it's ok to check for that exact version string

Nadrieril: Do you know how to do such a check in a makefile? I'd rather not run it on every test

msprotz: I would do something like

```
SHELL=/bin/env bash
_ := $(shell bash -c '(( ${BASH_VERSION%%.*} >= 4 ))')
ifneq ($(.SHELLSTATUS),0)
_: $(error "bash version too old")
endif
```

(sketch)

msprotz: probably need to use $$ though to prevent make from attempting variable expansion

PR #140, opened 2025-01-31
Refresh expected output

msprotz:
This should fix the eurydice build

### DIFF
--- a/out/test-array2d/array2d.h
+++ b/out/test-array2d/array2d.h
@@ -27,6 +27,9 @@ void array2d_main(void);
 extern bool
 core_cmp_impls___core__cmp__PartialEq_u32__for_u32__24__eq(uint32_t *x0, uint32_t *x1);
 
+extern bool
+core_cmp_impls___core__cmp__PartialEq_u32__for_u32__24__ne(uint32_t *x0, uint32_t *x1);
+
 #if defined(__cplusplus)
 }
 #endif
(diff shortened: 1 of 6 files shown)


out/test-array2d/array2d.h
```diff
@@ -27,6 +27,9 @@ void array2d_main(void);
 extern bool
 core_cmp_impls___core__cmp__PartialEq_u32__for_u32__24__eq(uint32_t *x0, uint32_t *x1);
 
+extern bool
+core_cmp_impls___core__cmp__PartialEq_u32__for_u32__24__ne(uint32_t *x0, uint32_t *x1);
+
```
Nadrieril, line 32: You might not be using the latest version of charon locally, this shouldn't be there.

msprotz:
Ok looks like we need both https://github.com/cryspen/libcrux/pull/782 and https://github.com/FStarLang/karamel/pull/527 to get kyber to pass. Wonder in what order I should merge these? @Nadrieril can you remind me what to do in these situations...? I forgot. Thanks!

Nadrieril:
Either way around is good I think. I'd probably merge eurydice first since eurydice is more central to libcrux's CI than vice-versa so it's more important to get the libcrux side right.

msprotz:
Great. So we override CI just for this one?

Nadrieril:
yep

msprotz:
temporarily changed the settings to allow administrators to bypass rules

PR #141, opened 2025-02-05
Better codegen for array repeats in Rust -- fixes #137

### LINKED ISSUE #137: Imperfect array initialisation
Initialising arrays is translated into a long list of assignments, which blows up the code size and may lead to bad compilation (depending on compiler and optimisation level).

```rust
fn init() -> [u8; 32] {
    [0u8; 32]
}
```

```c
void eurydice_tests_init(uint8_t ret[32U])
{
  ret[0U] = 0U;
  ...
  ret[31U] = 0U;
}
```

msprotz:


msprotz:
@franziskuskiefer I would like you to try this out on mlkem and mldsa before merging, so I'll wait until we get libcrux back to green before merging this, I'm mildly concerned with regressions

please also note that in the case of *nested* arrays, there will still be a long series of assignments -- if this is a problem I left a TODO in the code that indicates what to tweak to improve that situation, too

msprotz:
Note that this patch still improves things: previously, 

```rust
  17   │ fn init2() -> [[u8; 32]; 32] {
  18   │     [[0u8; 32]; 32]
  19   │ }
```

would generate 1024 assignments -- now it only generates 32, but with 32 distinct buffers on the stack...

msprotz:
ok looks like I need to revamp the array repeat desugaring phase

msprotz:
As far as I can tell, this generates a rather pleasant diff on branch protz_refresh of libcrux -- see my latest commit there to see the effect of this PR.

@franziskuskiefer if the changes are agreeable to you, please go ahead and merge... thanks

franziskuskiefer (changes requested):
Thanks for tackling this!

This improves a bunch of cases.

But there's at least one it makes worse.

```rust
let mut tmp_stack = [[0i32; 263], [0i32; 263], [0i32; 263], [0i32; 263]];
```

(Maybe this should be different in Rust already, but that's a different question.)
used to generate

```c
int32_t tmp_stack[4U][263U] = {{0U}};
```

Now it generates

```c
int32_t repeat_expression[263U] = {0U};
int32_t tmp_stack[4U][263U] = {
    {(int32_t)0, ...},
    {(int32_t)0, ...},
    {(int32_t)0, ...},
    repeat_expression};
```

Which is not only worse, but won't compile.

msprotz:
Ah good catch. I'll add this as a test-case, I didn't have it locally. Thank you

msprotz:
I ended up revamping the handling of nested arrays and simplified pretty awful code. (I actually vividly remember writing this code in a rush and thinking I wasn't sure I was doing the right thing...)

You'll need https://github.com/FStarLang/karamel/pull/530

The consequence should now be uniformly better code. For your specific example I do not yet generate a perfect nested initializer (i.e. `int32_t tmp_stack[4U][263U] = { 0 }`) but the code is pretty good. Let me know if you'd like me to push a little more on code quality.

For instance, if you have arrays nested three levels deep (i.e. arrays of arrays of arrays) there is an opportunity to introduce a for-loop in the code I just touched, rather than statically generating a slew of assignments. With the big cleanup, this should be pretty easy to do, so if this is something you need, let me know. Cheers.

msprotz:
I was able to confirm that on my branch protz_refresh of mlkem, I see no concrete impact, and this passes the tests locally.

franziskuskiefer (approved):
Thanks. Looks like this is good to go now.

PR #159, opened 2025-03-18
Add extern C to eurydice_glue.h for gcc C linkage

franziskuskiefer:
The core functions are declared as part of the extraction (C linkage) and conflict with the implementations in glue (potentially C++ linkage). Clang appears to be fine with this. GCC is not. For example

```
/home/runner/work/libcrux/libcrux/libcrux_core.h:53:24: error: conflicting declaration of ‘uint64_t core_num__u64_9__from_le_bytes(uint8_t*)’ with ‘C’ linkage
   53 | static inline uint64_t core_num__u64_9__from_le_bytes(uint8_t x0[8U]);
```

This PR adds extern C to most of the `core_*` functions to avoid this. That's not the best solution, but should fix the immediate build issue. Ideally we wouldn't generate the function declarations in the first place.

### DIFF
--- a/include/eurydice_glue.h
+++ b/include/eurydice_glue.h
@@ -199,6 +199,26 @@ static inline void Eurydice_slice_to_array3(uint8_t *dst_tag, char *dst_ok,
 
 // CORE STUFF (conversions, endianness, ...)
 
+static inline uint8_t Eurydice_bitand_pv_u8(uint8_t *p, uint8_t v) {
+  return (*p) & v;
+}
+static inline uint8_t Eurydice_shr_pv_u8(uint8_t *p, int32_t v) {
+  return (*p) >> v;
+}
+static inline uint32_t Eurydice_min_u32(uint32_t x, uint32_t y) {
+  return x < y ? x : y;
+}
+
+#define core_num_nonzero_private_NonZeroUsizeInner size_t
+
+// These need an extern C block because they are defined as extern C in the
+// extraction.
+// Note that clang is apparently ok with this conflicting declaration. But GCC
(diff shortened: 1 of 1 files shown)


msprotz:
Why does extraction have C linkage? I thought the -fc++17-compat flag got rid of `extern "C"` entirely in the code-gen.

msprotz:
Looks like the glue is going to need extern "C" (default) or no extern "C" (-fc++17-compat), so this is tricky to get right...?

franziskuskiefer:
The scenario is a C library, which is used from C++. In that case we include`*_core.h` in the C++ code, which has the function declarations inside `extern "C"` blocks. But the definition in the glue, does not have an `extern "C"`, hence gets included as C++.
I'd really like to avoid two different glue files. This was just to have something that works/shows the issue. I don't have a good idea for a nice solution here yet.

msprotz:
I thought about this and the best I can suggest is to have `KRML_EXTERN static void` ... for each declaration, then:

in eurydice_glue.h:

```c
#ifndef KRML_EXTERN
#define KRML_EXTERN extern "C"
#endif
```

in your build system, for C++:
`-DKRML_EXTERN`

in your build system, for regular C code, nothing

thoughts?

msprotz:
Addressed for all static inline stubs in https://github.com/AeneasVerif/eurydice/pull/166

@franziskuskiefer -- you will need to define KRML_CXX17_COMPAT (see diff over there) to assert that you're in the special mode where there is no extern "C" at all, and you use a C++ compiler, to entirely skip those definitions.

PR #164, opened 2025-03-27
Support repeat in toplevel expressions

msprotz:


msprotz:
ERROR: Eurydice expects Charon to be invoked with exactly --remove-associated-types '*'
       For full logs, run 'nix log /nix/store/psyzqyd7l2j09hky11pgdlyrw0pzxyjx-ml-kem.drv'.

PR #169, opened 2025-03-28
Also export rust toolchain in flake

Nadrieril:
Followup to https://github.com/AeneasVerif/eurydice/pull/168: I also need the rust toolchain.


Nadrieril:
Turns out `inputs.eurydice.inputs.charon` works well.

PR #170, opened 2025-03-31
cremepat: credible, meta-extensible patterns

msprotz:
With the proliferation of rewriting rules that aim to resugar, clean up and massage AST nodes, we got into a totally untenable situation where one had to write actual pattern matches in Eurydice that were 50+ lines. This is extraordinarily tedious, and not future-proof. See https://github.com/AeneasVerif/eurydice/blob/9c587e76b6724c4a4e887b4443327c5680544288/lib/Cleanup2.ml#L610-L663 for an example.

With this, we replace pattern matches on AST nodes with concrete syntax for krml's internal AST. Programmers can write concrete syntax, using `?x` to bind an (OCaml) pattern variable, or `?x..` to bind an (OCaml) pattern variable that has (OCaml type) list.

Programmers can also refer to bound variables in concrete syntax; this gets compiled to an OCaml pattern match that references the correct bound variable depending on how many binders were traversed. Of course, this is not entirely hygienic.

Internally, a lexer and parser build a syntax tree for the concrete syntax -- then, this gets compiled to an OCaml pattern-match, using some generic type combinators and polymorphic recursion to avoid duplication all the places in the AST where pattern or list pattern variables may appear.

As a proof of concept, I added more rewriting rules for loops that don't rely on their indices, and this took me a fraction of the time.


msprotz:
@Nadrieril can we get Dune 3.13? I need it for the 3.0 version of the Menhir facility inside of Dune. Thanks

msprotz:
@franziskuskiefer this should fix poor code quality for iterator-based loops that do not use their indices

Nadrieril:
@msprotz should work now (https://github.com/AeneasVerif/eurydice/pull/172).

msprotz:
Ok now it's missing menhir, let me try to fix this

msprotz:
@Nadrieril if you could just double-check that I did the right thing for menhir, that'd be great -- thanks

PR #188, opened 2025-05-13
Preliminary support for DSTs

msprotz:
This PR provides preliminary support for DSTs. For now, we are limited to:
- types of the form `T<U: ?Sized>` (single type argument)
- coercions from `T<[V; N]>` to `T<[V]>`

There are plenty of comments in the PR, but the short version is that we emit the following code:

```c
// T<[V;N]>
typedef struct {
  ... regular fields ...
  V data[N];
} T0;

// T<[V]>
typedef struct {
  ... regular fields ...
  V data[]; // C flexible array member
} T1;

// Box<T<[V; N]>>
T1*

// Box<T<[V]>>
typedef struct {
  T1 *ptr;
  size_t len;
} Eurydice_dst_T1;
```

The test should provide some clarity as to what is going on.


msprotz:
Thanks @Nadrieril -- appreciate you double-checking my compilation strategy!

PR #190, opened 2025-05-20
Sync Charon 690

N1ark:
Sync with https://github.com/AeneasVerif/charon/pull/690 -- I didn't add support for `CopyNonOverlapping`, as I'm not familiar with the codebase or how it works:/ Encountering it will just error, but adding support for it should be fairly simple given it's literally just `memcpy` with reordered args

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ out/*/a.out
 *.orig
 *.llbc
 eurydice
+_opam/*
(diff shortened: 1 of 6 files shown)


N1ark:
Ah oops, didn't mean to commit those -- just had 50k comitted files because of opam, along with a weird make error, irrelevant :p

Nadrieril:
I'm interested in the weird make error. We added these lines to the Makefile exactly to avoid weird make errors; if you know how we can do this better, I'm all ears

PR #192, opened 2025-05-21
A few tweaks; new field in config: add_to_internal_h

msprotz:
This might break the mlkem build since I changed the name of some macros...

### DIFF
--- a/lib/Builtin.ml
+++ b/lib/Builtin.ml
@@ -367,7 +367,7 @@ let unwrap : K.decl =
   let open Krml in
   let open Ast in
   let lid =
-    [ "core"; "result"; "{core::result::Result<T, E>[TraitClause@0, TraitClause@1]}" ], "unwrap"
+    [ "core"; "result"; "{core::result::Result<T, E>}" ], "unwrap"
   in
   let t_T = TBound 1 in
   let t_E = TBound 0 in
--- a/bin/main.ml
+++ b/bin/main.ml
@@ -94,12 +94,12 @@ Supported options:|}
       [
         [ "Eurydice" ], "slice_index";
         [ "Eurydice" ], "slice_subslice";
-        [ "Eurydice" ], "slice_subslice2";
+        [ "Eurydice" ], "slice_subslice3";
         [ "Eurydice" ], "slice_subslice_to";
         [ "Eurydice" ], "slice_subslice_from";
         [ "Eurydice" ], "array_to_slice";
         [ "Eurydice" ], "array_to_subslice";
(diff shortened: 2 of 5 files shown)


lib/Builtin.ml
```diff
@@ -367,7 +367,7 @@ let unwrap : K.decl =
   let open Krml in
   let open Ast in
   let lid =
-    [ "core"; "result"; "{core::result::Result<T, E>[TraitClause@0, TraitClause@1]}" ], "unwrap"
```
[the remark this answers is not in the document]
Nadrieril: what's the issue?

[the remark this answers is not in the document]
msprotz: I have an error in mlkem and I'm wondering if it could be the case that mlkem is running with an older charon that emits the previous name [TraitClause@0, ...

lib/AstOfLlbc.ml
```diff
@@ -2376,7 +2389,7 @@ let file_of_crate (crate : Charon.LlbcAst.crate) : Krml.Ast.file =
   } =
     crate
   in
-  if options.remove_associated_types <> [ "*" ] then begin
+  if options.preset <> Some (Eurydice) then begin
```
msprotz, line 2392: @Nadrieril ultimately this was the root cause -- the behavior changed between symcrust and libcrux because the former was not passing --preset=eurydice, and the comment below was updated but not the runtime check

msprotz:
Ok mlkem is going to need a toolchain upgrade for this.

> ml-kem> ERROR: Eurydice expects Charon to be invoked with `--preset=eurydice`

first error, then the second error is that because libcrux vendors its own copy of eurydice_glue.h, they do not yet have the Eurydice_subslice_to_array3 macro

lib/AstOfLlbc.ml
Nadrieril: I don't understand, the preset implies the `remove_associated_types` option, how could this have changed anything?

Nadrieril: I kept the runtime check the same on purpose to avoid breaking libcrux, this was a forward-compatible change to avoid disruption.

msprotz:
I had to leave some old workarounds in there to not break the libcrux build. @franziskuskiefer what I would need from libcrux is
1. invoke charon with --preset=eurydice
2. do a single invocation of charon instead of 3 currently (@Nadrieril you said once this should be possible, although I can't quite remember how to do this...?)
3. pass `--include core::num::*::BITS --include core::num::*::MAX` to charon

These three things would allow me to get rid of some backwards-compat stuff I left in this PR for libcrux only

PR #199, opened 2025-05-23
Add core_num__u64_9__rotate_left

franziskuskiefer:
Add `core_num__u64_9__rotate_left` to the glue.

cc https://github.com/cryspen/libcrux/issues/992


msprotz (approved):
Super thanks so much!

PR #200, opened 2025-05-28
Update charon

Nadrieril:
Companion to https://github.com/AeneasVerif/charon/pull/702.


lib/AstOfLlbc.ml
```diff
@@ -174,6 +174,7 @@ let string_of_path_elem (env : env) (p : Charon.Types.path_elem) : string =
         | ImplElemTrait _ -> "(" ^ i_as_str ^ ")"
       in
       "{" ^ i_as_str ^ d ^ "}"
+  | PeMonomorphized _ -> Charon.PrintTypes.path_elem_to_string env.format_env p
```
Nadrieril, line 177: @msprotz When you try charon's monomorphization, here you'll have access to the generic arguments used to monomorphize hte given item so you can reproduce your current name encoding scheme.

PR #204, opened 2025-05-30
Properly substitute method generics

Nadrieril:
This fixes method generics substitution by using the method binders as intended.

### DIFF
--- a/lib/AstOfLlbc.ml
+++ b/lib/AstOfLlbc.ml
@@ -995,37 +995,46 @@ let rec mk_clause_binders_and_args env ?depth (trait_clauses : C.trait_clause li
                     C.ConstGenericVarId.of_int
                       (C.ConstGenericVarId.to_int varid + ambient_ts.K.n_cgs)
                   in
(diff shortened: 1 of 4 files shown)


lib/AstOfLlbc.ml
```diff
@@ -922,70 +943,140 @@ let rec build_trait_clause_mapping env ?depth (trait_clauses : C.trait_clause li
...
+                in
+                { C.binder_params = bound_fn.binder_params; binder_value = signature }
+              in
+
+              (* We then construct a polymorphic signature for this method.
+                 Its generics are the method generics (the ones in the binder).
+                 *)
+              let method_sig =
+                Charon.Substitute.(
+                  (* Variables bound in the inner binder are `Bound`, which
```
msprotz, line 984: can you say more about what is not handled here, and where it should be handled?

msprotz (approved):
I think I roughly understand what's going on. Let's switch to Charon-based monomorphization asap this is getting way too complicated.

PR #208, opened 2025-06-05
opaque type translation

ssyram:
Originally, Eurydice simply drops the external types with private fields (opaque types), which causes many converted codes referring to undefined types. Thanks to the update in Charon in [Charon/PR#704](https://github.com/AeneasVerif/charon/pull/704). Now we can fetch the layout info and translate the opaque types to:

```C
typedef struct Type_s { uint8_t data[SIZE]; } Type;
```

Here `SIZE` is the actual size coming from the layout info.

This helps maintain the correct semantics of the original program and inter-crate modular translation.

Currently, the monomorphised types are still not supported to fetch layout info. Once Charon is ready, the code should work smoothly for types like `Vec<i32>`.

### DIFF
--- a/lib/AstOfLlbc.ml
+++ b/lib/AstOfLlbc.ml
@@ -2252,12 +2252,13 @@ let decl_of_id (env : env) (id : C.any_decl_id) : K.decl option =
            `typdef struct { uint8_t data[SIZE]; } Type;` *)
         match decl.layout with
         | Some { C.size = Some size; _ } ->
-          if const_generics <> [] || type_params <> [] then
-            fail "Internal error: the opaque type %a has layout, yet it is also generic."
-              plid name;
-          Some (K.(DType (name, [], 0, 0, Flat [
-            Some "data", (TArray (Krml.Helpers.uint8, (SizeT, string_of_int size)), false)
-          ])))
+          if const_generics <> [] || type_params <> [] then begin
+            L.log "AstOfLlbc" "The type %a has layout info but it is generic, dropped. Recommendation: generate the LLBC with `--monomorphize` option." plid name;
+            None
+          end else
+            Some (K.(DType (name, [], 0, 0, Flat [
+              Some "data", (TArray (Krml.Helpers.uint8, (SizeT, string_of_int size)), false)
+            ])))
         | _ ->
           if const_generics <> [] || type_params <> [] then
             L.log "AstOfLlbc" "The type %a is generic, hence no static layout info." plid name


ssyram:
Well, a more temporary way would be to parse the result from "-Zprint-type-sizes", I guess everything is already included.

Nadrieril:
Let's keep that discussion to the Zulip

msprotz:
Your proposal description is very high-level, so I'm going to try to fill in the blanks so that we can describe precisely what it is that you propose. If I had to guess, what you're proposing is the following. Please fill in the blanks and/or rephrase, I left question marks where I'm trying to guess.

Background:
- to support separate compilation, Rust provides size information for types defined in other crates (?) so as to allow stack-allocating them (??) -- this is e.g. for the case where a crate provides `pub fn f() -> T` but `T` is not `pub` (???)
- this allows reconciling at link-time code that was previously compiled separately
- this is known as "opaque types" and is different from https://rustc-dev-guide.rust-lang.org/opaque-types-type-alias-impl-trait.html which is another concept entirely (but also called opaque types)

(Question: is this only for a multi-crate scenario or can this happen within a single crate?)

Proposal:
- such opaque types compile to `typedef struct Type_s { uint8_t data[SIZE]; } Type;`
- this allows allocating proper size information
- C files can be compiled separately and reconciled at link-time, thus mimicking what is happening in Rust

(Question: what if only some of the fields of a type are private? what happens then?)

Open questions:
- the C compiler might introduce some extra padding but it's ok to have more size than needed
- as long as the two definitions (the complete definition and the opaque definition) do not coexist within the same C file we should be ok

Please comment -- thanks!

ssyram:
Thanks for the note. Translating everything very easily leads to errors when the codes use `std`.

Nadrieril:
In this case the easiest solution is to use `--include` and `--exclude` to select which definitions to extract. Hopefully like that you can avoid the definitions that charon can't extract for now.

msprotz:
@Nadrieril and I are discussing this right now -- we have further thoughts about your proposal:
- we are concerned it will lead to irreconcilable issues owing to the C ABI calling convention, as dictated by either Windows or the System-V ABI
- lacking alignment information, the sizes will not match
- this might lead to issues with strict aliasing violations.

Questions for you:
- how do you handle the case of alignment?
- how can you argue that the C compiler will generate the same code when there are precise rules that dictate whether passing struct by value happens in registers or by spilling on the stack? I thought these rules were type-directed, especially for arm64 where register vs. stack-passing depends on the presence of floats or not in the struct.

msprotz:
@Nadrieril's proposal seems like it might help you achieve what you want -- perhaps let us know a little bit more about your use-case? we might be able to help you better this way (and if this is something you'd rather discuss in private, that's fine, and we can have a private chat on zulip!) cheers

msprotz:
> Rust promotes whole-dependency compilation schemes where all the dependencies are available in cargo... what prevents you from doing a single run of Charon?

I also forgot to mention that --extract-opaque-bodies should pick all the definitions that are reachable

ssyram:
Thanks for the useful info! Let us study more on this case!

Issue #203, opened 2025-05-29
Reborrow leading to type error?


ssyram:
When I test:

```Rust
fn reborrow_rule() {
  let mut s = String::from_str("abc").unwrap();
  let r = &mut s;
  {
    let rr = &mut *r;
    println!("{rr}");
  }
  println!("r: {r}");
  println!("s: {s}");
}
```

The strange type-checking problem happened:
```
Warning 4: in the arguments to core.fmt.rt.{core::fmt::rt::Argument<0>#1}.new_display <  alloc_string_String* > [[core.fmt.{(core::fmt::Display␣for␣&0␣mut␣(T))#76}.fmt < alloc_string_String > [[alloc.string.{(core::fmt::Display␣for␣alloc::string::String)#23}.fmt]]]] [[alloc.string.{(core::fmt::Display␣for␣alloc::string::String)#23}.fmt]], in the sequence statement at index 11, after the definition of uu____49, in top-level declaration test_eurydice.features.reborrow_rule, in file test_eurydice: Malformed input:
subtype mismatch:
   alloc_string_String* ->  core_fmt_Formatter* -> core_result_Result () core_fmt_Error (a.k.a.  alloc_string_String* ->  core_fmt_Formatter* -> core_result_Result () core_fmt_Error) vs:
    alloc_string_String** (a.k.a.   alloc_string_String**)
```

It seems to be something to do with the type of the library type? I have no idea why this happens.

msprotz:
I think it has mostly to do with the incorrect compilation scheme for `str` which you intend to fix, corret?

ssyram:
No, I don't think this is the case -- this relates to the reborrowing of &String rather than &str.

msprotz:
I'm having a hard time reproducing your error. I can successfully extract your example to C, but then I have C compilation errors because the String type is not defined, and because the println! macros don't work. See below.

```
jonathan@Mac:~/Code/eurydice (main) $ git rev-parse head
2ca24a2e00b05a7eeef7431e37dd113411191f8b
jonathan@Mac:~/Code/eurydice (main) $ make test-reborrow^C
jonathan@Mac:~/Code/eurydice (main) $ cat test/reborrow.rs
───────┬────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
       │ File: test/reborrow.rs
[...]

PR #211, opened 2025-06-10
Add an extra test for #203

msprotz:


msprotz:
(to be merged before support for String comes in)

PR #214, opened 2025-06-11
fn-ptr casting

ssyram:
Originally, the casting handles only the case when the `from` and `to` types are equivalent. This is too strict for function pointer castings. As the lifetime info is erased in the C code. Now, if the functions types differ only in lifetime, the casting is allowed. The lifetime correctness is assumed to be provided by Rust. Moreover, to the best of my knowledge, I think Rust should always guarantee that function casting happens only when (1) they differ in only the lifetime variables; and (2) the lifetimes are unifiable. So no additional check should be required? Nevertheless, I added the check for robustness to avoid any unexpected error.

This newly resolves cases like the following:
```Rust
fn id_ref<'a, T>(x: &'a T) -> &'a T {
    x
}

fn applies<'a, 'b, T, R>(f : fn(&'a T) -> &'b R, arg : &'a T) -> &'b R {
  f(arg)
}

fn main() {
  assert_eq!(*applies(id_ref, &1), 1)
}
```

This example is added to the tests set.


lib/AstOfLlbc.ml
```diff
@@ -1604,6 +1604,17 @@ let is_str env var_id =
...
 let is_dst_var env var_id = Option.is_some (is_dst env (snd (lookup env var_id)))
 
+let rec differs_in_lifetime_only (f : C.ty) (t : C.ty) =
+  match f, t with
+  | C.TRawPtr (f, fk), C.TRawPtr (t, tk)
+  | C.TRef (_, f, fk), C.TRef (_, t, tk) -> fk = tk && differs_in_lifetime_only f t
+  | C.TArrow f, C.TArrow t ->
+    let f_args, f_ret = f.binder_value in
+    let t_args, t_ret = t.binder_value in
+    List.length f_args = List.length t_args &&
```
Nadrieril, line 1614: That check is not needed, `for<'a> fn(&'a u32, &'a u32)` and `for<'a, 'b> fn(&'a u32, &'b u32)` are inter-convertible.

Nadrieril:
It's a sanity check that cannot cause soundness issues because we have to trust rustc anyway, I think it's fine as it is.

Nadrieril:
I'm ok with that

PR #217, opened 2025-06-13
added floating point support

ssyram:
The support for `f32` and `f64` is added. This is not going to compile at the time, before Krml approves the addition of `Float32` and `Float64` as `width`. New examples added for floating points with check passed.


ssyram:
Now Krml should be ready? @protz

msprotz:
Yes can you merge main into your branch?

msprotz:
This is the error:        > Fatal failure in floating_points.main: Invalid_argument("is_signed")

ssyram:
No, sorry...

Issue #220, opened 2025-06-16
Feature: global const should not be macro


ssyram:
Although macro is a convenient mechanism in C, I think we should be conservative in translating Rust objects into it. Not only because macro loses the type information but also may cause incorrect codes. For example, if we have:

```Rust
const SOME_INT : &&i32 = &&0;
```

This cannot be properly translated to C -- currently fatal error and exits.

I think the proper translation should be:

```C
int32_t SOME_INT_deref_deref = 0;
int32_t *SOME_INT_deref = &SOME_INT_deref_deref;
int32_t **SOME_INT = &SOME_INT_deref;
```

Or, at least `static` should not be macros as they currently are.

PR #221, opened 2025-06-20
Revised `const` and `static` support

ssyram:
Linked to issue #220, previously, `const` & `static` declarations in Rust were all translated to C macros. In this case, definitions like:

```Rust
const C_VAL: &&i32 = &&0;
static S_VAL: &&i32 = &&1;
```

were translated to:
```C
#define C_VAL (&&0)
#define S_VAL (&&0)
```

which is not valid C code. Because `0` and `&0` are not valid lvalues in C.

Now, we fix this to be:
```C
void *C_VAL_local_0 = (void *)0U;

void **C_VAL_local_1 = &C_VAL_local_0;

void ***C_VAL_local_2 = &C_VAL_local_1;

#define C_VAL (&C_VAL_local_2)

int32_t S_VAL_local_0 = (int32_t)0;

int32_t *S_VAL_local_1 = &S_VAL_local_0;

int32_t **S_VAL_local_2 = &S_VAL_local_1;

int32_t ***const S_VAL = &S_VAL_local_2;
```

This translation takes care of the difference between `const` and `static` in Rust, as the `static` variables are lvalues in Rust while `const` variables are purely inlined rvalues.

The key change here includes:
- In `AstOfLlbc.ml`, we distinguish between `const` and `static` variables -- now only `const` is marked `Macro`.
- In `Cleanup2.ml`, we removed the addition of `const` flags to all global definitions in `build_macros`.
- We added the `globalise_global_locals` pass to `globalise` the local variables in the global variable definitions to also global variables to ensure the generated lvalues are valid.
- We debugged the `check_addrof` function to now recursively generate lvalues for nested expressions.
- `GlobalRef` has now been supported in `expression_of_rvalues` of `AstOfLlbc.ml`.


.vscode/tasks.json
Nadrieril, line 1: This file is not appropriate for eurydice

lib/AstOfLlbc.ml
```diff
@@ -1837,9 +1837,14 @@ let expression_of_rvalue (env : env) (p : C.rvalue) : K.expr =
...
         (TArray (typ_of_ty env t, constant_of_scalar_value (assert_cg_scalar cg)))
         (K.EBufCreateL (Stack, List.map (expression_of_operand env) ops))
-  | Global { id; _ } ->
+  | Global { id; _ } | GlobalRef ({ id; _ },_) -> begin
       let global = env.get_nth_global id in
-      K.with_type (typ_of_ty env global.ty) (K.EQualified (lid_of_name env global.item_meta.name))
+      let e = K.with_type (typ_of_ty env global.ty) (K.EQualified (lid_of_name env global.item_meta.name)) in
+      match p with
+      | Global _ -> e
+      | GlobalRef _ -> K.(with_type (TBuf (e.typ, false)) (EAddrOf e))
```
msprotz, line 1845: minor, but I would prefer if you had two separate cases rather than a nested pattern-match -- it would improve readability (and it's ok to duplicate just two lines of code, especially if they're next to one another)

lib/Cleanup2.ml
```diff
@@ -563,12 +563,9 @@ let build_macros (macros : Krml.Idents.LidSet.t ref) =
     inherit [_] map as super
 
     method! visit_DGlobal env flags name n t body =
-      if Krml.Helpers.is_bufcreate body then
-        super#visit_DGlobal env flags name n t body
-      else begin
-        (macros := Krml.Idents.LidSet.(union !macros (singleton name)));
-        DGlobal (flags @ [ Macro ], name, n, t, body)
-      end
+      if List.exists (fun x -> x = Krml.Common.Macro) flags then
```
msprotz, line 566: List.mem

lib/Cleanup2.ml
```diff
@@ -1132,3 +1131,98 @@ let inline_loops =
       else
         DFunction (cc, flags, n_cgs, n, t, name, binders, body)
   end
+
+(** A better version of hoist (than [Krml.Simplify.hoist]), also work for [DGlobal]. *)
```
msprotz, line 1135: would you prefer to implement this upstream in krml? would that make your life easier?

lib/Cleanup2.ml
```diff
@@ -1132,3 +1131,98 @@ let inline_loops =
...
+    become:
+    [T1 VAL_local_1 = e1;
+     T2 VAL_local_2 = e2[v1/VAL_local_1];
+     ...
+     TN VAL_local_N = eN[v1/VAL_local_1; v2/VAL_local_2; ...; vN-1/VAL_local_(N-1)];
+     T VAL = e;]
+
+     Notably, the locals should be renamed to avoid potential naming conflicts.
+*)
+let globalise_global_locals files =
```
msprotz, line 1187: `globalize`, please (most of krml is written in en-US, not en-GB)

lib/Cleanup2.ml
```diff
@@ -1132,3 +1131,98 @@ let inline_loops =
...
+  let mapper = function
+  | DGlobal (flags, name, n_cgs, ty, expr) ->
+    let rec decompose_expr id infoAcc expr =
```
msprotz, line 1190: nit: info_acc (snake_case for variable names)

msprotz (commented):
Ok, overall this looks good -- thanks!

I only had rather minor comments... I think we should be able to merge this easily once you fix the cosmetic issues. Thanks!

lib/AstOfLlbc.ml
ssyram: Resolved, the lines are long...

lib/Cleanup2.ml
ssyram: I think the current version is OK, as modifying an established project seems potentially problem-causing?

ssyram: Thanks!

ssyram:
I'm working on Mac and can only use `gmake`. The test does not seem to change anything...

msprotz:
I posted more detailed remarks on how to regenerate the contents of out/ properly in #231, but let's keep the main discussion here.

There is one remaining problem with your patch (which you can see in #231 since it has a clean diff): it sometimes introduces an un-necessary level of indirection, see for instance https://github.com/AeneasVerif/eurydice/pull/231/files#diff-a264507e0c1360e808c97f3d246af7f9696c2c8d7692520cb711e55f967a8274R89-R91 -- I don't think this diff should be there.

This is undesirable because it changes the name of the symbol in the object file. Can you please fix this last detail? Thank you!

ssyram:
I see. But I think that indirection is correct as they follow the principles discussed above: const itself should be a pure right-value. And if the value it is assigned to in Rust is not a pure right-value, we should introduce the left-value.

msprotz:
Let me take a look and see if I can debug this

PR #227, opened 2025-07-02
Closure to fn casting (by bug fix for single tuple)

Lin23299:
Given the update on closure in Charon, the casting from closure to fn comes almost for free.
I've made 2 changes in  'expression_of_rvalue' in AstOfLlbc.ml and added this test case:

```
fn main(){
    let f: fn(i32) -> i32 = |x| x + 42;
    f(1);
}
```

### 1 CastFnPtr 
It now only allows casting from FnPtr and FnDef(which includes the closure case) to FnPtr like follows.
Other cases will fail. I think it is more robust than the previous solution for CastFnPtr.

```
  | UnaryOp (Cast (CastFnPtr (TFnDef _, TFnPtr _to)), e) ->
      let dst = typ_of_ty env (TFnPtr _to) in
      K.with_type dst (K.ECast (expression_of_operand env e, dst))
    (* possible safe fn to unsafe fn, same in C *)
  | UnaryOp (Cast (CastFnPtr (TFnPtr _, TFnPtr _)), e) ->
     expression_of_operand env e
```

### 2 The bug on single tuple?

The following code misses the case for single tuple [op],  this actually fails the test case in the current version. (the casting of closures with more than 1 argument are fine)

```
  | Aggregate (AggregatedAdt ({ id = TTuple; _ }, _, None), ops) ->
      let ops = List.map (expression_of_operand env) ops in
      let ts = List.map (fun x -> x.K.typ) ops in
      if ops = [] then
        K.with_type TUnit K.EUnit
      else begin
        assert (List.length ops > 1);
        K.with_type (TTuple ts) (K.ETuple ops)
```

I simply added the case `[op] -> expression_of_operand env op` and the test case works.

Is this actually a bug or it's designed to fail the `assert` for single tuple?


lib/AstOfLlbc.ml
```diff
@@ -1810,13 +1810,14 @@ let expression_of_rvalue (env : env) (p : C.rvalue) : K.expr =
...
-        assert (List.length ops > 1);
-        K.with_type (TTuple ts) (K.ETuple ops)
+      begin
+        match ops with
+        | [] -> K.with_type TUnit K.EUnit
+        | [op] -> expression_of_operand env op
+        | _ ->
+           let ops = List.map (expression_of_operand env) ops in
+           let ts = List.map (fun x -> x.K.typ) ops in
+           K.with_type (TTuple ts) (K.ETuple ops)
```
Nadrieril, line 1818: That surprises me, @msprotz is there special-casing of single-element tuples elsewhere in eurydice/krml?

test/closure_fn_cast.rs
```diff
@@ -0,0 +1,4 @@
+fn main(){
+    let f: fn(i32) -> i32 = |x| x + 42;
+    f(1);
+}
```
Nadrieril, line 4: Could you also commit the generated C code like for the other tests?

lib/AstOfLlbc.ml
```diff
@@ -1713,6 +1713,13 @@ let expression_of_rvalue (env : env) (p : C.rvalue) : K.expr =
   | UnaryOp (Cast (CastRawPtr (_from, to_)), e) ->
       let dst = typ_of_ty env to_ in
       K.with_type dst (K.ECast (expression_of_operand env e, dst))
+    (* from FnDef to FnPtr *)
+  | UnaryOp (Cast (CastFnPtr (TFnDef _, TFnPtr _to)), e) ->
```
msprotz, line 1717: cosmetic: comments referring to a particular branch are located immediately underneath the pattern arm elsewhere in the codebase (so, move the comment one line below)

msprotz (commented):
Great PR! Thanks for this first contribution.

Looks good, like @Nadrieril said you need to commit the test output so that i) we can see the effect of your patch on the PR and ii) so that you can pass CI... thanks!

Lin23299:
Thank you! I've added the test output and fixed the code style. Hopefully this can pass CI :)

Nadrieril (approved):
Looks good for my side of things

msprotz:
Looks almost perfect. Congrats!

Do you think you could add a special-case so that we do not emit a cast when source and destination types are identical? It would generate cleaner code. I think it's simply a matter of doing `if _from = _to then expression_of_operand e else (* your new code *)`

PR #229, opened 2025-07-02
Formatting

msprotz:


msprotz:
@Nadrieril wondering if there's a way to run `make format-check` as part of the CI... (assuming everyone can easily install the right version of ocamlformat, of course)

PR #234, opened 2025-07-08
Revised version of PR #221

msprotz:
This is a followup fix on top of @ssyram's #221 

The two main changes in this (in addition to what @ssyram did already) are:
1. Use https://github.com/FStarLang/karamel/pull/617 to minimize the amount of code duplication (using an OCaml class instead of an anonymous object)
2. Reinstate special treatment that was needed for toplevel array declarations. My comment in the most recent commit should be pretty explanatory, but the short version is that if something has an array type at toplevel, then it is not eligible for being emitted as a macro. Previously, the code was aware of this, but @ssyram's changes removed that special treatment, which in turn required rewriting the top-level declaration to emit additional bindings... my tweak fixes this.

A minor change is that this PR contains a refreshed out/ directory which minimizes the diff.


ssyram:
Thanks for this! The fix in Krml is definitely good (given that it does not bring unintended change in its original purpose)! While for the second point, thanks for pointing out the subtlety but I am still thinking adding these additional rules for arrays may break the theoretical integrity... It would be clean and predictable that all `const` is turned to macro and all `static` is turned to global variable, with inevitable implicit locals in globals globalised. This rule should be more explicit and easier to reason about?

ssyram:
OK, that makes sense.

PR #237, opened 2025-07-10
Code quality improvement

msprotz:
This PR contains the following code quality improvements:
- aggressively inline variables named left_val and right_val, on the basis that these are temporaries emitted by assert! and debug_assert!
- mark more functions as side-effect free, to facilitate their inlining and thus improve code quality
- rewrite try_with_capacity in abstract syntax and keep the macros to a minimum
- retain `meta` data on *binders* in let-bindings, which in turn allows keep the `AttemptInlining` information for left_val and right_val, above
- improve array cleanup phases to be aware that structs are acceptable array initializers, and that recursively, array literals can be used to initializer fields of structs that have array types (instead of turning these into assignments, later on to be desugared using `memcpy`s)
- optimization in the handling of Rust repeat expressions


msprotz:
@Nadrieril as discussed: I implement the functions we needed either using abstract syntax (tedious, but manageable) or using macros

I'd be interested in your review to make sure my error cases for try_with_capacity are in sync with what Rust does (I'm not sure what the layout information is used for, but I hope I put stuff that makes sense in there)

Nadrieril:
What I can see from reading the rust implementaiton of `try_with_capacity`:
- I suspect your `try_with_capacity` doesn't handle ZSTs correctly (a `Vec<()>` should not call `malloc` I believe, depends what malloc says about zero-sized allocations);
- The length logic is a bit more complex: instead of just doing `len * sizeof(t)`, the distance between array elements must be the smallest multiple of the alignment that's bigger than `sizeof(t)`; e.g.

```rust
// This means that every `S` must be stored at an address multiple of 8.
#[repr(align(8))]
struct S(u8);
```

- The `layout` returned should be the layout of the data we tried to initialize, i.e. with size `len * size` (using the size mentioned just above).

PR #245, opened 2025-07-18
Support for discriminant reads, as discussed in aeneasverif/charon#774

Nadrieril:


msprotz:
But I'm in favor of merging now and adjusting later -- this should not be a big deal to fix once I have a good test case

Issue #37, opened 2024-07-01
subtype mismatch in black_box


franziskuskiefer:
The following code 

```rust
fn b(x: &[u8]) -> [u8; 32] {
    [0u8; 32]
}

pub fn bb(x: &[u8]) -> [u8; 32] {
    core::hint::black_box(b(x))
}
```

produces the following error when running it through eurydice

```
subtype mismatch:
  uint8_t[32size_t] -> uint8_t[32size_t] (a.k.a. uint8_t[32size_t] -> uint8_t[32size_t]) vs:
  uint8_t[32size_t] -> uint8_t[32size_t] -> () (a.k.a. uint8_t[32size_t] -> uint8_t[32size_t] -> ())
```

msprotz:
This feels almost identical to the fix I just pushed for slice_index -- https://github.com/AeneasVerif/eurydice/pull/93

(note to self for when I investigate this further)

msprotz:
So the issue is that black_box is an external function, and as such will be emitted as a macro. Macros only have one signature, and for `black_box`, the signature of the macro is `<T> T -> T`.

Remember that Rust arrays are values, and generate copies; however, they are compiled as `T[N]` in C. This means we rewrite types, so that a function type that takes and returns an array (like `black_box`) becomes a function type that takes a source and a destination array, and returns unit.

This means that there is a mismatch because the call-site wants a function of type `uint8_t[32size_t] -> uint8_t[32size_t] -> ()` (after monomorphization and type rewriting) but black_box is an external that keeps the type `<T> T -> T`, which instantiates to uint8_t[32size_t] -> uint8_t[32size_t] in this case.

There are several options here:
1. monomorphize external function prototypes -- for instance, Eurydice might emit references to `black_box_u8[32]`, `black_box_[u8]`, and potentially a lot more variants to account for each type that `black_box` is used at -- then, the user would have potentially infinitely many macros to "fill out" to provide implementations
2. monomorphize external function prototypes, but only for arrays -- this means that there would be two variants of black_box generated internally; one would be `black_box<T> (dummy: T) -> T` and the other would be `black_box_array<T> (dummy: T) (dst: T) -> ()`, with the invariant that the latter is used only at call-sites that (after monomorphization) instantiate black_box with an array type
[...]

Issue #212, opened 2025-06-11
Bug: Slice references translation error from generics


ssyram:
Currently, `&[T; N]` is translated to `T*` while `[T; N]` is translated to `T [N]`, which makes sense and works fine in most of the cases. But problem happens when `&[T; N]` results from generics types application, in this case, it results in `T [N] *`. The following codes produce this error:

```Rust
fn id_mut<'a, T>(x : &'a T) -> &'a T { x }

fn use_id_mut() {
  let arr : [i32; 3] = [1,2,3];
  let r : &[i32; 3] = id_mut(&arr);
  assert!(r[0] > 0);
}
```

The error is as expected:
```
subtype mismatch:
   int32_t* (a.k.a.  int32_t*) vs:
   int32_t[3size_t]* (a.k.a.  int32_t[3size_t]*)
```

Btw, shall we get to the monomorphized LLBC asap? I understand that problems like this happens due to the indirectness from generics. Monomorphized codes will definitely help getting rid of this layer of indirectness.

ssyram:
Sure, thanks!

ssyram:
One more reason of supporting PR #252 by @Lin23299 . With the PR, this bug is now fixed!

msprotz:
PR #252 should also fix #37 I believe

ssyram:
Yes, that is the key motivation of #252 !

PR #252, opened 2025-07-25
Translating rust array type [T;C] into struct arr { T[C] data; } 

Lin23299:
According to our discussion in https://github.com/AeneasVerif/eurydice/issues/37, I've done with this prototype which can pass `make test`. The cases are mainly unchanged except adding some `assert!` to avoid unused variable warning.

I'll introduce the main changes with my thoughts on them. This version surely needs more cleanup but maybe you can take a look first and give some feedback :)

### Translation
In `AstOfLlbc.ml`, three basic changes:

1. The types of arrays are translated into the following type: 
```ml
let arr : K.lident = [ "Eurydice" ], "arr"
let mk_arr (t : K.typ) (cg: K.cg) : K.typ = K.TCgApp (K.TApp (arr, [ t ]), cg)
```

2. The corresponding `decl_of_Arr` is included in the translated program.
```ml
let decl_of_Arr =
  K.DType (Builtin.arr , [], 1, 1, Flat [(Some "data", (K.TCgArray (TBound 0,0), true))])
```
3. The expressions of arrays come from `[x;N]` or `[x,y,z]` and functions related to arrays are similarly translated.

### Cleanup phases
In `Cleanup2.ml`:
  1. In `remove_implicit_array_copies`, a recursive call for multi-demensional array is added.
  2. The translation of `from_fn` and `map` functions are changed from the pass-by-ref style to directly returning the struct of array as value.
  3. `remove_array_repeats` is changed by 1) find and translate the `arr{T data[N];}` pattern in expressions; 2) force `visit_DGlobal` to return a initializer list as macro definitions in C.

A new phase `remove_builtin_decls` in `Cleanup3.ml` is added to remove several specific monomorphized array types which are already defined in `eurydice_glue.h` for builtin functions (e.g. `[u8;4]` for `core_num__u32__to_be_bytes`).

### External definitions

As we discussed in https://github.com/AeneasVerif/eurydice/issues/37, I changed the function types in `Builtin.ml` to use the new struct types. `box_new_array` and `slice_of_boxed_array` are no longer needed.

Notably, most of the C macro definitions of functions for slices and arrays in `eurydice_glue.h` are unchanged. Because it seems that the pointer to the struct is  nearly "the same" as the pointer to the array `data`. It works for now and I'm not sure should we keep this or add more type-polymorphic macros as @msprotz mentioned.

### Others

As mentioned in https://github.com/FStarLang/karamel/issues/627, I've added a new builtin function 'empty_array' to replace the empty buffer list for 0-length array. it's later implemented as follows:
```c
#define Eurydice_empty_array(dummy, t, t_dst)                                  \
  ((t_dst) { { } })
```
I still need a dummy arg to pass the checker, maybe you have better idea?


lib/AstOfLlbc.ml
```diff
@@ -493,10 +506,11 @@ and maybe_cg_array env t cg =
 
 (* To be desugared later into variable hoisting, allocating suitable storage space, followed by a
    memcpy -- this is just a placeholder and isn't even type-checked. *)
-let mk_deep_copy (e : K.expr) (l : K.expr) =
+
+(*let mk_deep_copy (e : K.expr) (l : K.expr) =
```
msprotz, line 510: remove

lib/AstOfLlbc.ml
```diff
@@ -702,10 +716,10 @@ let rec expression_of_place (env : env) (p : C.place) : K.expr =
       (* L.log "AstOfLlbc" "e=%a\nty=%s\npe=%s\n" pexpr sub_e (C.show_ty sub_place.ty) *)
       (*   (C.show_projection_elem pe); *)
       match pe, sub_place, sub_place.ty with
-      | C.Deref, _, TRef (_, TAdt { id = TBuiltin TArray; generics = { types = [ t ]; _ } }, _)
+      (*| C.Deref, _, TRef (_, TAdt { id = TBuiltin TArray; generics = { types = [ t ]; _ } }, _)
```
msprotz, line 719: please remove the code that is commented out -- no reason to keep it if it belongs to an old compilation scheme that is now gone... it'll also make the diff easier to review... thanks!

lib/AstOfLlbc.ml
```diff
@@ -714,7 +728,7 @@ let rec expression_of_place (env : env) (p : C.place) : K.expr =
           (* All types represented as a pointer at run-time, compiled to a C pointer *)
           begin
             match !*sub_e.K.typ with
-            | TBuf (t_pointee, _) | TArray (t_pointee, _) ->
+            | TBuf (t_pointee, _) | TArray (t_pointee, _) -> (**??*)
```
msprotz, line 731: not sure why you're putting (**??**) here, but if I had to guess, the TArray case does not make sense; if so, maybe put `| TArray _ -> assert false` for now, and we'll remove it in a little while once we've confirmed that this causes no regressions

lib/AstOfLlbc.ml
```diff
@@ -1582,14 +1596,19 @@ let expression_of_fn_ptr env (fn_ptr : C.fn_ptr) = expression_of_fn_ptr env "" f
 
 let expression_of_operand (env : env) (op : C.operand) : K.expr =
   match op with
-  | Copy p ->
-      let e = expression_of_place env p in
+  | Copy p -> expression_of_place env p
+     (** is this necessary? if all the types of array are already translated into struct in [typ_of_ty],
```
msprotz, line 1600: agreed that this now can go

lib/AstOfLlbc.ml
```diff
@@ -1730,63 +1756,26 @@ let expression_of_rvalue (env : env) (p : C.rvalue) expected_ty : K.expr =
...
-                    _;
-                  };
-              },
-            TAdt
-              {
-                id = TBuiltin TBox;
-                generics =
-                  { types = [ TAdt { id = TBuiltin TSlice; generics = { types = [ t2 ]; _ } } ]; _ };
-              } ) ->
-            (* Cast from Box<[T; N]> to Box<[T]> which we represent as Eurydice_slice.
```
msprotz, line 1766: can you explain why this case is gone?

lib/Cleanup2.ml
```diff
@@ -209,7 +208,7 @@ let remove_array_repeats =
              let the subsequent `hoist` phase lift this into a let-binding, and code-gen will optimize this into
              { 0 }. If we can't do something smart, we let-bind, and fall back onto the general case. *)
           begin
-            try (self#expand_repeat false (with_type (snd env) (EApp (e, es)))).node
+            try (self#expand_repeat false (fst env) (with_type (snd env) (EApp (e, es)))).node
```
msprotz, line 211: can you explain what this boolean stands for somewhere?

lib/Cleanup3.ml
```diff
@@ -159,6 +159,22 @@ let add_extra_type_to_slice_index =
       | _ -> super#visit_EApp env e es
   end
 
+(*This identifies the decls which should be generated after monomorphism, but is already defined
+ in eurydice_glue.h for implementing the builtin functions. Currently only for arr<T;N> *)
+let is_builtin_lid lid = match lid with
```
msprotz, line 164: if you wanted, you could customize lib/krml/Monomorphization.ml to pick readable names in a few specific cases, such as when the type argument if of the form `[T; N]` (and use the hash otherwise)

out/test-nested_arrays2/nested_arrays2.h
```diff
@@ -15,7 +15,23 @@
...
+
+/**
+A monomorphic instance of Eurydice.arr
+with types Eurydice_arr uint8_t[[$1size_t]]
+with const generics
+- $1size_t
+*/
+typedef struct Eurydice_arr_81_s { Eurydice_arr_f1 data[1U]; } Eurydice_arr_81;
+
+#define NESTED_ARRAYS2_TABLE ((KRML_CLITERAL(Eurydice_arr_81){ .data = { (KRML_CLITERAL(Eurydice_arr_f1){ .data = { 1U } }) } }))
```
msprotz, line 34: surprised this comes back as a define -- is your branch up to date with main?

out/test-slice_array/slice_array.c
```diff
@@ -16,38 +16,57 @@ _uint8_t__x2;
...
 }
 
 void slice_array_f2(void)
 {
-  uint8_t x[4U][4U] = { { 0U } };
+  Eurydice_arr_11
+  x =
+    {
+      .data = {
+        (KRML_CLITERAL(Eurydice_arr_e9){ .data = { 0U, 0U, 0U, 0U } }),
```
msprotz, line 48: it would be nice, in the case of nested initializers like this, to *not* print the KRML_CLITERAL -- it's not necessary

out/test-where_clauses_fncg/where_clauses_fncg.c
```diff
@@ -16,9 +16,9 @@ with const generics
 - K= 8
 - L= 4
 */
-uint64_t where_clauses_fncg_bar_16_7b(uint8_t x[8U][4U], uint8_t _[4U][8U])
+uint64_t where_clauses_fncg_bar_16_7b(Eurydice_arr_cb x, Eurydice_arr_1c _)
 {
-  return (uint64_t)x[0U][0U];
+  return (uint64_t)x.data->data[0U];
```
msprotz, line 21: I do not understand why there is a -> here

msprotz (commented):
Thanks for this PR! Overall, this looks good, and I'm quite pleasantly surprised by the improvements in code quality: the struct declarations are more verbose, but the upside is that there are much fewer mempcy's all over the place. So I find your PR quite convincing!

I have reviewed things at a high level, and I have not reviewed cleanup2.ml in detail yet. I have a few requests before I start doing an in-depth review:
1. please remove old code, rather than commenting it out -- it makes the diff easier to review
2. please explain how you managed to eliminate the case of the DST conversion from Box<[T;N]> to Box<[T]> -- I think this is showcased in test-castunsize, but a comment would help
3. please think about avoiding undefined behavior for 2. -- right now, the cast to `cast_unsize_T *` violates strict aliasing -- I don't have a good answer for this, and this issue was there before your PR, but I feel like your PR makes it slightly worse because it now casts across very different structs (as opposed to casting between two structs only the last field of which differs), and there might be issues related to alignment
4. can you clarify what happens with empty arrays? I feel like you're triggered undefined behavior as well (see https://en.cppreference.com/w/c/language/struct.html "Similar to union, the behavior of the program is undefined if struct is defined without any named members (including those obtained via anonymous nested structs or unions). ")
5. can you please checkout libcrux on branch protz/ci and run ./c.sh in libcrux-ml-kem, followed by `cd c/ && cmake -B build && cmake --build build` and then run the tests? I remember that the cleanup2 phase was quite delicate to get right and libcrux was a very good test for that feature

Thanks! Looking forward to taking a more in-depth look at this

ssyram:
Example for #212 has been added to the test case, the bug of which has been resolved by this PR!

lib/AstOfLlbc.ml
```diff
@@ -383,6 +387,19 @@ let lookup_field env typ_id field_id =
...
 
+
+(** special treatment for the array type: translating [T;C] as rust generic type
+    struct <const C:usize, T> { data : [T;C]; } *)
+
+let decl_of_Arr =
+  K.DType (Builtin.arr , [], 1, 1, Flat [(Some "data", (K.TCgArray (TBound 0,0), true))])
+  (* []  : no flags
+     1  : we have one const generic C
+     0  : we have one type argument T
```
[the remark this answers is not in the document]
Lin23299: fixed

include/eurydice_glue.h
```diff
@@ -540,6 +553,10 @@ static inline char *malloc_and_init(size_t sz, char *init) {
 #define Eurydice_box_new(init, t, t_dst)                                       \
   ((t_dst)(malloc_and_init(sizeof(t), (char *)(&init))))
 
+// Initilize the struct Eurydice_arr with an empty array as its field
```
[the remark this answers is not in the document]
msprotz: we may have to do that -- let's wait until the PR is updated and other questions are addressed, and then we can have maybe a special type for zero-sized arrays (something like `typedef struct { char data[1]; } Eurydice_zero_arr`)

lib/AstOfLlbc.ml
```diff
@@ -55,6 +55,10 @@ type var_id =
   | ConstGenericVar of C.const_generic_var_id
   | Var of C.local_id * C.ety (* the ety aids code-generation, sometimes *)
 
+(* The type of obligation for extra declarations needed during the translation
+   The minimal necessary information should be stored in the env and extended to
+   complete declaration at the end*)
+
```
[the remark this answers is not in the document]
Lin23299: It's just legacy from the previous implementation, removed.

lib/AstOfLlbc.ml
Lin23299: Done

Lin23299:
> 5. can you please checkout libcrux on branch protz/ci and run ./c.sh in libcrux-ml-kem, followed by `cd c/ && cmake -B build && cmake --build build` and then run the tests? I remember that the cleanup2 phase was quite delicate to get right and libcrux was a very good test for that feature

I cannot find branch named "protz/ci" in [eurydice](https://github.com/AeneasVerif/eurydice) and [libcrux](https://github.com/cryspen/libcrux). Which version should I test on?

msprotz:
My bad -- branch `protz/ci` is on my fork https://github.com/protz/libcrux/tree/protz/ci here. Thanks for testing!

One thing I do when I test these things is I first extract + test using Eurydice main, then I add the output C code under version control (`git add + git commit`), then I re-run extraction using the new branch of Eurydice (in this case, array_handling), then I use `git diff` to easily see what was impacted.

Once you confirm this is good, I'll also measure the impact of the changes on https://github.com/AeneasVerif/symcrust. Then, the remaining steps will be:
- figure something out for UB (casts + zero-sized arrays)
- in-depth review of the changes to cleanup2

msprotz:
https://github.com/protz/libcrux/tree/protz/ci-arr contains fixed for the topological sort issue. I also added the original C code under version control so that we can see via `git diff` the impact of the changes.

However, there is an error further down that I'm trying to minimize -- for some reason, a function is not monomorphized.

msprotz:
I also started seeing these errors, maybe it's related:

```
thread 'rustc' panicked at compiler/rustc_middle/src/ty/context.rs:2946:9:
assertion failed: eps.array_windows().all(|[a, b]|
        a.skip_binder().stable_cmp(self, &b.skip_binder()) !=
            Ordering::Greater)
warning: Hax panicked when translating `core::iter::traits::collect::IntoIterator`.
 --> /rustc/library/core/src/iter/traits/collect.rs:282:1
```

pausing debugging work on this for now while I wait for @Nadrieril to get back to me... thanks!

Lin23299:
> 2. As far as our concerns, the violating strict alias rule might lead to the aligning problems. So the key is that we need to guarantee the correctness of the alignment shared across the pointed structures. Interestingly, we found that flexible array member `T data[]` has the **same** alignment of `T`, thus the same as `T data[N]` for some `N`, hence also the same as `Arr<T, N>` in the current PR. Hence, we believe it should be safe to simply take the address of the original `S<[T; N]>` for `S<[T]> *` in C. So here the only place to fix becomes that the `Eurydice::derefed_slice` should be a generic that turns to `T []` instead of `char []` now.

Based on this point, I've tried to generate a generic `T[]` from Eurydice_derefed_slice.  The problem is that current `Ast` can not express the *flexible array field* in dst struct.
The current version simply turns a field with type `Eurydice::derefed_slice T` into `Eurydice_derefed_slice` by omitting the type argument in `AstToCStar`.

A solution in my mind is to add a generic type for such flexible array field in krml such that we get it from `Eurydice_derefed_slice T` and translate it into `T[]` in C. I don't how hard it is to implement this. Otherwise we can define a generic type for `derefed_slice` similar as array, but the problem of translating it into `T[]` is still unsolved...

lib/Cleanup2.ml
```diff
@@ -95,11 +95,10 @@ let remove_implicit_array_copies =
                         (self#remove_assign n (lift lifting_index lhs_i) (lift lifting_index e)
                            (nest lifting_index (array_index + 1) es))
                   | _ ->
-                      with_type e2.typ
-                        (ELet
-                           ( H.sequence_binding (),
-                             H.with_unit (EAssign (lift lifting_index lhs_i, lift lifting_index e)),
-                             nest (lifting_index + 1) (array_index + 1) es )))
+                      with_type e2.typ (self#visit_ELet ((),TUnit)
```
msprotz, line 98: why do we need to go through visit_ELet below again?

msprotz: the intent of this code (as I recall it) is that remove_assign is not supposed to recurse back into the visitor -- it's for base cases of desugaring assignments

lib/Cleanup2.ml
```diff
@@ -236,9 +240,14 @@ let remove_array_repeats =
...
        committed to an initializer list (and Charon will suitably "fold" repeat expressions
        automatically for us), so we might as well expand.
+
+       Another boolean flag [under_global] is used to indicate we're translating a DGlobal -- where
+       we should also use an initlializer list. Maybe it should be unified with [under_bufcreate].
     *)
-    method private expand_repeat under_bufcreate e =
+    method private expand_repeat under_bufcreate under_global e =
       match e.node with
+      | EBufCreateL (l, es) ->
```
msprotz, line 249: I think this was explicitly avoided (see comment at line 190) because `[[1; 1]; 256]` would generate an enormous amount of code. Let's wait until we have libcrux CI back to see what the effect of this is.

lib/Cleanup2.ml
```diff
@@ -367,43 +383,63 @@ let remove_array_from_fn files =
...
             L.log "Cleanup2" "%a %a" ptyp t_elements ptyp t_elements;
-            (* First argument = closure, second argument = destination. Note that
-               the closure may itself be an application of the closure to the state
-               (but not always... is this unit argument elimination kicking in? not
-               sure). *)
-            let state, dst =
+            (* By translating array into struct, it is simply as return value of [from_fn]
+               instead of the 2nd arg as its reference. *)
+            let state =
               match es with
```
msprotz, line 389: Krml.KList.one

lib/Cleanup2.ml
```diff
@@ -367,43 +383,63 @@ let remove_array_from_fn files =
...
-                H.zero_usize (* i: size_t = 0 *),
-                H.mk_lt_usize (Krml.DeBruijn.lift 1 len) (* i < len *),
-                H.mk_incr_usize (* i++ *),
-                let i = with_type H.usize (EBound 0) in
-                Krml.Helpers.with_unit
-                  (if H.is_array t_dst then
-                     EApp
-                       ( call_mut,
+            (* Constructing types, maybe there are some better helpers for this idk *)
+            let t_struct =
```
msprotz, line 394: let _, t_struct = Krml.Helpers.flatten_arrow e.typ in

lib/Cleanup2.ml
```diff
@@ -367,43 +383,63 @@ let remove_array_from_fn files =
...
+              match e.typ with
+              | TArrow (_, _to) -> _to
+              | _ -> assert false
+            in
+            let t_element =
```
msprotz, line 399: same

lib/Cleanup2.ml
```diff
@@ -367,43 +383,63 @@ let remove_array_from_fn files =
                instead of having type size_t -> t_element, has type size_t -> t_element ->
```
msprotz, line 383: is this comment still true?

lib/Cleanup2.ml
```diff
@@ -367,43 +383,63 @@ let remove_array_from_fn files =
...
+                ( Krml.Helpers.fresh_binder ~mut:true "i" H.usize,
+                  H.zero_usize (* i: size_t = 0 *),
+                  H.mk_lt_usize (Krml.DeBruijn.lift 1 len) (* i < len *),
+                  H.mk_incr_usize (* i++ *),
+                  let i = with_type H.usize (EBound 0) in
+                  Krml.Helpers.with_unit
+                    (if H.is_array t_dst then
+                     (* note: this nested array case needs to be changed? not sure. *)
+                      (EApp
+                        ( call_mut,
```
msprotz, line 425: how do you know that call_mut is not evaluated multiple times?

lib/Cleanup2.ml
```diff
@@ -367,43 +383,63 @@ let remove_array_from_fn files =
...
+              match call_mut.typ with
+              | TArrow (_, (TArrow (_, _to))) -> _to
+              | _ -> assert false
+            in
+            let t_array =
+              match len.node with
+              | EConstant c -> TArray (t_element, c)
+              | _ -> assert false
+            in
+            let x, dst_struct = H.mk_binding ~mut:true "arr_struct" t_struct in
```
msprotz, line 409: I think I get what this code is doing, but shouldn't it go through the removal of array assignments, above?

msprotz (commented):
ok I looked at your code a little bit for Cleanup2 and I there are some things I don't understand / am unsure about -- let's try to address those and look at the effect of these changes on the code of libcrux, I remember that a lot of this code was fine-tuned to avoid some bad corner cases that generated ugly code

ssyram:
Thanks for the detailed response! After discussion, we think in most modern compilers, both the 0-len array field & pointer casting from `struct { char foo; int data[3]; }` to `struct { char foo; int data[]; }` are common features that are widely used in modern compilers & also industrial practices (like in our cases).

Also, if we are to use `memcpy` to do that, there is chances where the semantics is wrong:
```
let mut dst : Dst<[i32; 3]> = Dst { foo: 'a', dst: [1, 2, 3] };
{
    let dst_1: &mut Dst<[i32]> = &dst;
    dst_1.dst[0] = 0;
}
```
Then, if we do `memcpy`, `dst`'s field will not be changed. The whole stuff is WRONG. We suggest Eurydice to explicitly state to disable the strict aliasing rule --- this rule makes C unable or at least very hard to convert. To be honest, the industrial application (like in our case & Linux), the `--fno-strict-aliasing` option is always on. Linus also argued about the "insane"ness of the strict aliasing rule.

On the other hand, we also understand your concern about the C standard. We propose Eurydice to make a "--standard-c" option (can be a future work or waiting for some issues on this), which generates codes that stick to the C standard more strictly. But there will be cases where some Rust features cannot be expressed, like the above case... Also, there are further deviation from run-time memory layout difference. While we have the best code quality (minimal length & best clarity & best recovery of Rust semantics) as the default option, while requiring some additional options to support this should be reasonable.

lib/Cleanup2.ml
Lin23299: Sorry I cannot get the point here, why `call_mut` could be evaluated multiple times? The construction of for loop using `call_mut` is basically the same as the previous implemention. We only added a declaration of the struct, use the for loop to initialize it and finally return it.

lib/Cleanup2.ml
Lin23299: Can not get the point neither... Similarly the main change we made here is to declare a struct and "return" it (replacing the function call 'core::array::map') after initializing it using the for loop which is basically unchanged. Doing the removal of array assignments after removal of `from_fn` and `map` seems reasonable to me.

lib/Cleanup2.ml
Lin23299: This is a mistake. I added this case to provide the functionality for the following code in `visit_DGlobal` :
https://github.com/AeneasVerif/eurydice/blob/eefbfdf482d5e8b5568f25d0e2754f04f7f62b75/lib/Cleanup2.ml#L273-L281
for the newly added case:
https://github.com/AeneasVerif/eurydice/blob/eefbfdf482d5e8b5568f25d0e2754f04f7f62b75/lib/Cleanup2.ml#L301-L302
in order to correctly translate `repeat.rs`.

So a quick fix I've made is to add `when under_global` to this case. I think this solution is still Ad hoc since the 2nd special case for `visit_DGlobal` is not applied for the new case for struct of array. I'm not sure will this cause some inconsistency or bugs. Or maybe we can just *modify* the old cases since all declarations of arrays are not translated to structs of arrays. I'm not quite sure yet.

lib/Cleanup2.ml
msprotz: ok I had not expanded the diff above -- my concern was if `call_mut` was of the form `f ()` which would result in multiple evaluations of it because it now in a `for`-loop, but clearly here it's a pointer to the `call_mut` implementation of the trait, so it's always going to be a value.

lib/Cleanup2.ml
msprotz: this was more of a general comment about whether there was an opportunity for code-sharing, but since this code was in there before, let's leave it as-is

perhaps there is a way to extract these implementations of map and from_fn rather than have built-in implementations (which are harder to maintain), but we can do that later

msprotz:
Ok this PR is getting close to landing -- thanks for working with me on improving everything. Here is what remains.
1. I ran some extra regressions (currently not under CI, but important anyway). I ran `make transpile` on branch `protz/c` of https://github.com/AeneasVerif/symcrust -- this works on main but with your PR I get thing kind of errors -- can you investigate?
```
Cannot re-check symcrust.error.fmt_da as valid Low* and will not extract it.  If symcrust.error.fmt_da is not meant to be extracted, consider marking it as Ghost, noextract, or using a bundle. If it is meant to be extracted, use -dast for further debugging.

Warning 4: in the arguments to core.fmt.rt.{core::fmt::Arguments<'a>}.new_v1 😱 [[1size_t]] [[1size_t]], after the definition of args, in top-level declaration symcrust.error.fmt_da, in file symcrust_mlkem: Malformed input:
subtype mismatch:
   Eurydice_arr_80[[$1size_t]]* ->  Eurydice_arr_1d[[$1size_t]]* -> core_fmt_Arguments (a.k.a.  Eurydice_arr_80[[$1size_t]]* ->  Eurydice_arr_1d[[$1size_t]]* -> core_fmt_Arguments) vs:
   Eurydice_arr_7d* ->  Eurydice_arr_d4* -> core_fmt_Arguments (a.k.a.  Eurydice_arr_7d* ->  Eurydice_arr_d4* -> core_fmt_Arguments)
--------------------------------------------------------------------------------

Cannot re-check symcrust.ntt.poly_element_decode_and_decompress as valid Low* and will not extract it.  If symcrust.ntt.poly_element_decode_and_decompress is not meant to be extracted, consider marking it as Ghost, noextract, or using a bundle. If it is meant to be extracted, use -dast for further debugging.

Warning 4: in the arguments to symcrust.ntt.slice_to_sub_array 😱 [[4size_t]], in the definition of uu____1853, in the then branch, after the definition of n_bits_in_coefficient, in the then branch, in the sequence statement at index 0, after the definition of uu____1886, in the last element of the sequence, after the definition of n_bits_in_accumulator, in top-level declaration symcrust.ntt.poly_element_decode_and_decompress, in file symcrust_mlkem: Malformed input:
subtype mismatch:
  Eurydice_slice uint8_t -> size_t -> Eurydice_arr_d8[[$4size_t]] (a.k.a. Eurydice_slice uint8_t -> size_t -> Eurydice_arr_d8[[$4size_t]]) vs:
  Eurydice_slice uint8_t -> size_t -> Eurydice_arr_e9 (a.k.a. Eurydice_slice uint8_t -> size_t -> Eurydice_arr_e9)
```

2. Blocked on https://github.com/AeneasVerif/charon/issues/790 to assess the effect of this PR on libcrux
3. Understanding whether the code can be simplified in https://github.com/AeneasVerif/eurydice/pull/252#discussion_r2250940672
4. Filing followup issues for UB related to DSTs, so that we don't forget about it

I think we're getting close. Thanks!

lib/Cleanup2.ml
Lin23299: Sure I'll try to find a clean solution and test it. After fixing the new bug in `symcrust`.

lib/Cleanup2.ml
```diff
@@ -362,48 +379,50 @@ let remove_array_from_fn files =
...
+              let lift1 = Krml.DeBruijn.lift 1 in
+              with_type TUnit (EFor
+                ( Krml.Helpers.fresh_binder ~mut:true "i" H.usize,
+                  H.zero_usize (* i: size_t = 0 *),
+                  H.mk_lt_usize (Krml.DeBruijn.lift 1 len) (* i < len *),
+                  H.mk_incr_usize (* i++ *),
+                  let i = with_type H.usize (EBound 0) in
+                  Krml.Helpers.with_unit
+                    (if H.is_array t_dst then
+                     (* note: this nested array case needs to be changed? not sure. *)
```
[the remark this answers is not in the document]
Lin23299: When I change the code based on a test case, this case for nested array was not triggered. So I've not look into it. I'll try to figure it out with maybe some new examples.

ssyram:
> I'm not arguing about this -- believe me, this makes me very unhappy too. There is no good solution. I think the correct fix for now is to add somewhere in the documentation a bit of text that says that usage of DSTs requires -fno-strict-aliasing or all bets are off. Perhaps, in addition to that, we could emit static_assert(EURYDICE_ALLOW_NON_STANDARD_C) in any file that contains a usage of DSTs, therefore requiring the user to opt-in.

Cool! I think that's a trade-off to make things in Rust better working in C. And your proposal is definitely a good solution to warn the users of this fact!

Lin23299:
Thanks for the replies! I've found the problem in:

> 1. I ran some extra regressions (currently not under CI, but important anyway). I ran `make transpile` on branch `protz/c` of https://github.com/AeneasVerif/symcrust -- this works on main but with your PR I get thing kind of errors -- can you investigate?

It seems that it is related to the monomorphization of the generic type `Eurydice_arr<T,C>` when it is used as argument of an *opaque* function like `new_v1` from [core::fmt::rt](https://doc.rust-lang.org/stable/src/core/fmt/rt.rs.html) . Another buggy function `symcrust.ntt.slice_to_sub_array` is marked as `#[charon::opaque]`, I tried to remove this mark and the errors for it disappear.

This error can be reproduced using a single `println!` which calls `new_v1`:
```rust
fn main() {
  println!("{}","Anything");
}
```
Before monomorphization (i.e. Phase 2.12), the type of `new_v1` is
```
ARRAY:
  type  Eurydice_arr t0 =<cg: 1> flat { mutable data: 0[$0] }
  external core_fmt_rt_{core::fmt::Arguments<'a>}_new_v1 <0>:<cg: 2>
    size_t ->
    size_t ->

        Eurydice_arr
        Eurydice_str[[$1]]* ->

          Eurydice_arr
          core_fmt_rt_Argument[[$0]]* ->
          core_fmt_Arguments
```
After it (i.e. Phase 2.13) it get translated to:
```
ARRAY:
  type auto_generated  Eurydice_arr_80  =<cg: 0>
    flat
    { mutable data: Eurydice_str[$0] }

  type auto_generated  Eurydice_arr_1d  =<cg: 0>
    flat
    { mutable data: core_fmt_rt_Argument[$0] }

  external core_fmt_rt_{core::fmt::Arguments<'a>}_new_v1 <0>:<cg: 2>
    size_t ->
    size_t ->

        Eurydice_arr_80[[$1]]* ->

          Eurydice_arr_1d[[$0]]* ->
          core_fmt_Arguments
```
Note the the array struct is "partly" monomorphized with only type but not cg. Meanwhile, the actual arguments used to call `new_v1` are typed using fully monomorphized types:
```
  type auto_generated  Eurydice_arr_7d  =<cg: 0>
    flat
    { mutable data: Eurydice_str[1size_t] }

  type auto_generated  Eurydice_arr_7b  =<cg: 0>
    flat
    { mutable data: core_fmt_rt_Argument[0size_t] }
```
Which causes the type mismatch when calling `new_v1` although they are actually the same type. I tried to release the checker, then the `AstToCStar` complains about unsugared `TCgArray`. So it seems to me that we should not generate types like `Eurydice_arr_80` and `Euryduce_arr_1d`.

This was not a problem on `main` because these array types are decayed to the same pointer. I've not looked deeply into `Monomorphization.ml`, but in this special case I think we should monomorphize `new_v1` (generally other opaque generic functions) according to the types of arguments when it is called instead of doing such "partly monomorphize" things.
What do you think? If agreed, do you have any suggests on how to fix it?

ssyram:
Thanks for the detailed discussion! After a careful discussion, we find emitting the external functions as their monomorphised version is a better choice. I would like to provide a more general point of view from the angle of the overall framework while @Lin23299 will provide some more technical obstacles about the original plan, which hence serves as evidence to support our claim.

Current mechanism requires manual provision of the macro implementation. This somehow violates the automatic principle of Eurydice --- we hope to extract C codes automatically. Currently, Eurydice monomorphises internal generic functions, and ultimately "delete" the generic prototype while leaving only the monomorphised versions, this is reasonable for C. But this also means these functions are not emitted to the outside for use by other libraries that monomorphise the functions with their own types. So, this creates a gap. We think a more unified framework would be better. Previously, I proposed a modular compilation proposal, but both you and @Nadrieril proposed to use the mechanism of Charon to extract the correct codes when needed.

Also, in the future, if we are transferring to the mono codes by Charon, we will also not get any generic codes any more. Everything, including the external will be all monomorphised. To get compatible with this method, it would also be good to simply monormorphise the external functions.

lib/Cleanup2.ml
Lin23299: I've changed these part of code quite a lot. In summary:

1. keep this recursive case for `EBufCreateL` in `expand_repeat` (only activated when we are comitted to initializer list);
2. remove the first two cases in `visit_EGlobal`, only leaving the call to `expand_repeat` with `under_global` flag on. The special case for `[c;n]` where `c` is a constant but not 0 is moved to `expand_repeat`;
3. Unified the two flags `under_bufcreate` into `under_global` since the former is only set to true in `visit_EGlobal`

Reasons:
1. Adding this recursive case for `EBufCreateL` in `expand_repeat` is a more complete implementation than just unfolding it for 1 time in `visit_EGlobal`.
For example: 
```rust
const ARRAY_3D : [[[u8;3];2];2] = [[
   [
     0,0,0,
   ], 
   [
     0,1,2
   ],
   ];2];
```
is translated into the shape of `ArrayRepeat -> BufCreateL -> ArrayRepeat`. The `expand_repeat` with the recursive case for `BufCreateL` can handle this. A similar implement like in the main cannot do so. 
Btw, `main` also fails on this case when translating it into a *const* instead of *macro*. Maybe the reason is similarly this recursive case, but I'm not sure.

2. Similarly, the special case of `[c;n]` defind in `expand_repeat` now can be triggered as a base case for nested array. It was only used for 1darrays.
3. The flag `under_bufcreate` was set to `true` only in `visit_DGlobal`. The point here is we commit to initializer list when translating a gloabl array which will become macros in C. So I think these two flags can be merged.

If you agree, I still have an uncertain thing here. Now this flag `under_global` is passed to the environment to recognize that we are committed to initializer list. Is this necessary? Is the current implementation functionally correct? I still need to find out, or maybe you have some advice?

msprotz:
> I've tested a bit on these code. Eurydice_arr_80[[$1size_t]] is indeed missing in the MonomorphizationTable. I have no idea on how to fix this.

Let me look. Thanks.

msprotz (commented):
Ok those changes look good. I added a bunch of cases in test/repeat.rs and they seem to make sense. There's one opportunity for simplification, perhaps -- see my comment. Thanks!

lib/Cleanup2.ml
msprotz: So overall it looks good to me -- does that answer your question?

msprotz:
@Lin23299 your analysis is correct, a partially-monomorphized type should never have been emitted. There was a corner case that the current test was not catching -- this is addressed by https://github.com/FStarLang/karamel/pull/636

Separately, I managed to run Eurydice with your branch on symcrust -- see https://github.com/AeneasVerif/symcrust/commit/e13f2b0e1fdfbcb5e909bb5f4fc6a9dffcbfa5d4 for the outcome... nothing terribly shocking in there

lib/Cleanup2.ml
Lin23299: Yes! Thanks for your reply.

lib/Cleanup2.ml
```diff
@@ -226,74 +226,70 @@ let remove_array_repeats =
...
-                    n,
-                    t,
-                    with_type e1.typ
-                      (EBufCreate (Stack, init, Krml.Helpers.mk_sizet (self#assert_length len))) )
-            | _ -> super#visit_DGlobal env flags name n t e1)
-        end
-      | _ -> super#visit_DGlobal env flags name n t e1
+      | EFlat [ (lido, _) ] when lido = Some "data" && self#is_arr_typ e1.typ ->
+          DGlobal (flags, name, n, t, (self#expand_repeat true) e1)   
+      | _ -> super#visit_DGlobal true flags name n t e1
```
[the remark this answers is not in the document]
Lin23299: True.. I've removed the `EFlat` case. It now only set the `env` to `true`.

msprotz:
Looks like the only thing that remains is to wait for @Nadrieril to be back from vacation and check that libcrux still works, correct?

msprotz:
Other than that I think we are in good shape with this PR -- I particularly appreciated your contributions in Cleanup2.ml and you are correct that this simplifies the translation greatly, so I'm all in favor of merging (once we get a green on libcrux).

Nadrieril:
It seems this is waiting on me, could you tell me what's needed from me?

msprotz:
In order to keep this PR moving, I re-generated the code in libcrux with this PR. Please see the diff here:

https://github.com/cryspen/libcrux/commit/c03a2450e05a21ae0aa53a715add84a7b759c4f4

Overall, things look pretty good! I'm rather pleased with the outcome -- some parts are more verbose (more struct definitions), but some parts are less verbose (less memcpy's all over the place).

@franziskuskiefer heads-up that this is the kind of diff that may land very soon, let's try to talk this week about whether you want to ride this change as part of your current upgrade of the toolchain, or whether you'd rather wait -- in any case, I'm not merging this without a plan to upgrade libcrux too

@ssyram @Lin23299 there is only one thing that I consider a code quality regressions: look at https://github.com/cryspen/libcrux/commit/c03a2450e05a21ae0aa53a715add84a7b759c4f4#diff-1f321a4519646efeacac463058a1ca42ce4884b9ce4169246258cfa2772e67f4R1866 -- now we have extremely long initializer lists. I think in krml, we need to apply the same initializer list shortening logic to compound literals, right now, it only applies to struct declarations as stack variables. I don't think it's a huge issue and you should be able to fix it quickly.

There is one other thing that I think is minor, and I'm not even sure we should fix it. Right now, if you do `type T = [u32; 2]` followed by `struct Foo { x: T }`, we generate `typedef struct { uint32_t data[2]; } Eurydice_arr_b2;` followed by `typedef Eurydice_arr_b2 T;`. On the one hand, it would be nice to be able to generate `typedef struct { uint32_t data[2]; } T;` directly, but on the other hand, this becomes unpredictable if there are several aliases for `[u32; 2]` (which name do we pick?). Happy to hear thoughts on this.

msprotz:
> This link does not work very well...

This is the diff -- file libcrux_mlkem_portable.c... there's more zeroes but you get the idea:

```diff
diff --git a/libcrux-ml-kem/c/libcrux_mlkem_portable.c b/libcrux-ml-kem/c/libcrux_mlkem_portable.c
index 3c073de99..713b30285 100644
@@ -2023,16 +1860,160 @@ libcrux_ml_kem.hash_functions.portable.shake128_squeeze_first_three_blocks with
 const generics
 - K= 4
 */
-static KRML_MUSTINLINE void shake128_squeeze_first_three_blocks_ac(
-    libcrux_ml_kem_hash_functions_portable_PortableHash_44 *st,
-    uint8_t ret[4U][504U]) {
-  uint8_t out[4U][504U] = {{0U}};
+static KRML_MUSTINLINE Eurydice_arr_ec
+shake128_squeeze_first_three_blocks_ac(Eurydice_arr_180 *st) {
+  Eurydice_arr_ec out = {
+      .data = {
+          (KRML_CLITERAL(Eurydice_arr_b0){
+              .data =
+                  {0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U,
+                   0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U,
+                   0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U,
+                   0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U,
+                   0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U,
+                   0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U,
+                   0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U,
+                   0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U,
+                   0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U,
+                   0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U,
+                   0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U,
+                   0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U,
+                   0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U,
+                   0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U,
+                   0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U,
+                   0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U,
+                   0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U,
+                   0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U,
+                   0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U,
+                   0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U,
+                   0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U,
+                   0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U,
+                   0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U,
+                   0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U,
+                   0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U,
+                   0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U,
+                   0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U,
+                   0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U,
+                   0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U,
+                   0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U,
+                   0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U,
+                   0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U,
+                   0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U,
+                   0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U}}),
```

> We believe that the current translation is better because of its consistency

Agreed.

msprotz:
Actually I'm surprised to see nested KRML_CLITERAL's -- this should not be needed.

Lin23299:
> I think in krml, we need to apply the same initializer list shortening logic to compound literals, right now, it only applies to struct declarations as stack variables. 

As you said, I've tried to add the case for compound literals as follows:
```
diff --git a/lib/CStarToC11.ml b/lib/CStarToC11.ml
index 4ea57007..c669a96d 100644
--- a/lib/CStarToC11.ml
+++ b/lib/CStarToC11.ml
@@ -239,6 +239,7 @@ let rec is_zero = function
 let rec trim_trailing_zeros l =
   match l with
   | Initializer [] -> Initializer []
+  | InitExpr (CompoundLiteral (_, l)) -> Initializer (List.map trim_trailing_zeros (chop_zeros l))
   | InitExpr e -> InitExpr e
   | Designated (d, init) -> Designated (d, trim_trailing_zeros init)
   | Initializer l -> Initializer (List.map trim_trailing_zeros (chop_zeros l))
```
> Actually I'm surprised to see nested KRML_CLITERAL's -- this should not be needed.

Note that thie line also *eliminates* the KRML_CLITERAL by turning `CompoundLiteral` into `Intializer`.

This makes the following difference in the test cases (they still pass):
```
diff --git a/out/test-slice_array/slice_array.c b/out/test-slice_array/slice_arr
ay.c
index 281b70c..614cec0 100644
--- a/out/test-slice_array/slice_array.c
+++ b/out/test-slice_array/slice_array.c
@@ -18,14 +18,7 @@ void slice_array_f1(void)
 {
   Eurydice_arr_11
   x =
-    {
-      .data = {
-        (KRML_CLITERAL(Eurydice_arr_e9){ .data = { 0U, 0U, 0U, 0U } }),
-        (KRML_CLITERAL(Eurydice_arr_e9){ .data = { 0U, 0U, 0U, 0U } }),
-        (KRML_CLITERAL(Eurydice_arr_e9){ .data = { 0U, 0U, 0U, 0U } }),
-        (KRML_CLITERAL(Eurydice_arr_e9){ .data = { 0U, 0U, 0U, 0U } })
-      }
-    };
+    { .data = { { .data = { 0U } }, { .data = { 0U } }, { .data = { 0U } }, { .data = { 0U } } } };
```
I'm a little bit uncertain about the elimination of `CompoundLiteral`, everything else seems good.
I'm going to create a PR for krml, maybe you can test it on the libcrux :)

msprotz:
Looks great. I think the "too smart" comment refers to exactly what you describe, where it gets a little confusing because some fields of the compound literal get omitted, for instance this would emit

```
typedef struct {
  int x, y;
} t;

...

t foo = { .x = 2 }
```

where field `y` is assumed to be zero -- I just think it's not very readable to omit `y`, which is why I disabled this behavior

franziskuskiefer:
> This is the diff -- file libcrux_mlkem_portable.c... there's more zeroes but you get the idea:

Ugh, that's pretty ugly. We worked before to get rid of these. I'd very strongly prefer not to get anything like this back. This blows up the code _a lot_. For mlkem for example this creates files that mostly consists of 0s for `Default` initialisation of arrays.

Do I understand correctly that Eurydice would generate APIs now that uses Eurydice defined types?
If that's right, I think that's a bad idea. This changes the entire integration story. Where we could say that consumers don't have to worry about these types before, they now have to construct them. The only way I can see that working is when "constructors" are provided for these types as well.

ssyram:
Yes, we are working on fixing these zeros.

franziskuskiefer:
> If that's right, I think that's a bad idea.

Actually, we have that in some places already. We just haven't integrated them yet. So it may not be that bad.

msprotz:
> Ugh, that's pretty ugly. 

I can confirm that most of the zeroes are gone.

There are only two instances that remains (they're identical):

```diff
@@ -3310,8 +3292,17 @@ with const generics
 - PARALLEL_LANES= 1
 - RATE= 136
 */
-void libcrux_sha3_generic_keccak_xof_zero_block_35_c6(uint8_t ret[136U]) {
-  memset(ret, 0U, 136U * sizeof(uint8_t));
+Eurydice_arr_3d libcrux_sha3_generic_keccak_xof_zero_block_35_c6(void) {
+  return (KRML_CLITERAL(Eurydice_arr_3d){
+      .data = {0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U,
+               0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U,
+               0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U,
+               0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U,
+               0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U,
+               0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U,
+               0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U,
+               0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U,
+               0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U}});
 }

 /**
```

But this can be easily fixed with:

```diff
diff --git a/lib/CStarToC11.ml b/lib/CStarToC11.ml
index c669a96d..b186fa23 100644
--- a/lib/CStarToC11.ml
+++ b/lib/CStarToC11.ml
@@ -1166,7 +1166,7 @@ and mk_compound_literal m name fields =
       ])
   | _ ->
       (* TODO really properly specify C's type_name! *)
-      CompoundLiteral (([], Named name, Ident ""), fields_as_initializer_list m fields)
+      CompoundLiteral (([], Named name, Ident ""), List.map trim_trailing_zeros (fields_as_initializer_list m fields))

 and struct_as_initializer m = function
   | Struct (_, fields) ->
```

And this removes even more zeroes than before.

msprotz:
We are still blocked on https://github.com/AeneasVerif/eurydice/pull/252#issuecomment-3138052351 -- @ssyram do you think that this issue is fixed by aeneasverif/charon#790?

msprotz:
If you have time, do you think you could look into this bug? It seems to be an x64-only behavior (and, specifically, on OSX): in those conditions, the function keccakf1600 is not extracted. `hax` also throws a lot of panics along the lines of:

```
thread 'rustc' panicked at compiler/rustc_middle/src/ty/context.rs:2946:9:
assertion failed: eps.array_windows().all(|[a, b]|
        a.skip_binder().stable_cmp(self, &b.skip_binder()) !=
            Ordering::Greater)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
warning: Hax panicked when translating `core::iter::traits::collect::IntoIterator`.
 --> /rustc/library/core/src/iter/traits/collect.rs:282:1
```

which might be related.

ssyram:
OK, so the main problem now is the sorting issue AeneasVerif/charon#790 ? This I think has been fixed by my PR in AeneasVerif/hax#2 . Yes, I think you're right that this is the discrepancy between the architecture... This is caused by that the order hash of Rustc used to implement `rustc_middle::ty::stable_cmp` function in Rustc produces a contradictory (while stable) result in the orders. So I fixed this by adding a sorting operation before calling the `mk_poly_existential_predicates` function.

ssyram:
OK, thanks!

protz:
Do you see hax in Cargo.lock for Charon? 

```
jonathan@MacBookPro:~/Code/charon (main) $ nix flake update hax --extra-experimental-features nix-command --extra-experimental-features flakes --override-input hax "github:aeneasverif/hax/f6ef814"
warning: input '' has an override for a non-existent input 'hax'
warning: the flag '--override-input hax github:aeneasverif/hax/f6ef814' does not match any input
warning: 'hax' does not match any input of this flake
```

this is the error I got

franziskuskiefer:
> Here: https://github.com/cryspen/libcrux/tree/protz/ci-arr

The `c` directory on that branch doesn't build for me unfortunately because there appear to be neon types in the core.

Trying to extract (`boring.sh`) I'm seeing a couple cycles such as the following. This means we have to explicitly put all array types into the config? Since there are platform specific pieces here I don't think we can use globs.
```
Eurydice.arr_51 (found in file libcrux_mlkem_core) mentions libcrux_ml_kem.vector.avx2.SIMD256Vector (found in file libcrux_mlkem768_avx2)
```

protz:
I was able to confirm that this does not break libcrux (see my latest commit on protz/ci-arr branch of libcrux). I'm going to try to figure out the CI issues and then merge this.

protz:
Congratulations! This has now been merged as part of #283. I will review the metadata PR tomorrow.

Thanks for all the good work and sorry it took so long to merge -- I wanted to be absolutely sure that nothing was broken.

PR #256, opened 2025-08-06
Re-instate #255

msprotz:
Looks like bad merge resolution on my end -- my bad.

@Lin23299 do you have a testcase for this?

### DIFF
--- a/lib/PreCleanup.ml
+++ b/lib/PreCleanup.ml
@@ -49,7 +49,7 @@ let remove_array_eq = object
         in
         assert (t = u);
         if is_flat t then
-          let diff = n_cgs - n_binders in
+          let diff = n_binders - n_cgs in
           match impl with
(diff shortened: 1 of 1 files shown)


Lin23299:
Testcase:
```
fn compare <const K:usize> (x : [u8;K], y : [u8;K]) -> bool {
  x == y
}
fn main() {
  let x = [2u8;2];
  let y = [2u8;2];
  let b = compare(x,y);
  assert_eq!(b,true);
}
```
But this testcase cannot find this bug on `main` branch, only make the difference in  #252 . Since this bug's effect is that the `CgVar` in signature of `array_eq`  goes wrong thus may mismatch with the its argument. However, if the array type `[T;C]` is decayed into `T*`, The cg `C` does not matter anymore.

Shall I add this into the test case `array.rs`?

PR #258, opened 2025-08-08
Fix eurydice_glue.h rotate_left

franziskuskiefer:
The implementation of `core_num__u64__rotate_left` has undefined behaviour for `x1 = 0`.


msprotz (approved):
Nice catch, thanks!

PR #259, opened 2025-08-09
Explicit Unsized Metadata & Unified DST Handling

ssyram:
I've implemented on the Charon side (AeneasVerif/charon#797) a functionality of making the pointer-metadata explicit in Charon. This reflects in that both `RawPtr` and `RvRef` in the `rvalue` type now will take an additional operand, which stores the metadata. So that, now, codes like:

```Rust
let dst : &[i32] = &x.dst;
```

Now will have the RHS represented as:
```OCaml
@9 := ptr_metadata(copy (x@1))
@3 := &((*(x@1)).dst, move (@9))
```

This is a part of the full DST (dynamically sized types) support plan. A new type `struct Eurydice::DstRef<Pointee, Meta> { Pointee *ptr, Meta meta }` is also added as the ultimate representation of such types. A function `mk_reference` is added, this function judges if the given metadata is `()`, the default case indicating no metadata is required (`Thin` pointer), otherwise, wrap the whole value with `Eurydice::DstRef`.

This has 3 major implications:
1. The whole workflow for slices should be restructured, as a slice `&[T]` is now simply `Eurydice::DstRef<T, usize>`, we may add an abbreviation as `type Eurydice::Slice<T> = Eurydice::DstRef<T, usize>` for better readability, though. This involves also the casting part in `AstOfLlbc` and many Cleanups.
2. The `str` type should now simply `Eurydice::DstRef<char, usize>`.
3. The original mechanism for handling "generic-DST" (i.e., types of the form `Type<U : ?Sized>`) can be now more generally handled. This should also make the codes more structured & clean.

Let us merge #252 first to partly alleviate the situation!

Of course, if we are simply to adapt to the new AST from Charon, we can ignore the final Operand for metadata, and everything will remain exactly the same as before.

### DIFF
--- a/libcrux-Cargo.lock
+++ b/libcrux-Cargo.lock
@@ -13,6 +13,31 @@ dependencies = [
 ]
 
 [[package]]
+name = "aes"
+version = "0.8.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b169f7a6d4742236a0a00c541b845991d0ac43e546831af1249753ab4c3aa3a0"
+dependencies = [
+ "cfg-if",
+ "cipher",
+ "cpufeatures",
+]
+
+[[package]]
+name = "aes-gcm"
+version = "0.10.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "831010a0f742e1209b3bcea8fab6a8e149051ba6099432c8cb2cc117dec3ead1"
+dependencies = [
+ "aead",
+ "aes",
+ "cipher",
+ "ctr",
+ "ghash",
+ "subtle",
+]
+
+[[package]]
 name = "aho-corasick"
 version = "1.1.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


msprotz:
Your plan seems good. I'm hoping to merge #552 shortly as soon as I confirm no breakages. Thanks.

Lin23299:
I shall first try to see the amount of work and do we need the cremepat PPX rewriter :)

ssyram:
@msprotz We have tested your suggestions with `--include "core::convert::_" --include "core::array::_"`, but the resulting stuff is quite complex and even involve the vtable, which is not yet matured (Hope AeneasVerif/charon#762 could get merged so we have a better stuff). I think before transiting to mono, for this specific case to work, we'd better implement by hand (by a pass).

msprotz:
Ok that sounds good to me then. Thanks for trying!

protz:
Ok there's only this one left -- please merge main into this branch to make sure there's nothing important left on this branch.

@Lin23299 congratulations on a super nice piece of work!

Lin23299:
Thank you very much! These instructions works for me and everything is pretty clean now. :)

PR #261, opened 2025-08-14
Update following changes in Charon

sonmarcho:


msprotz (approved):
Ok I see the pretty-printing of the `{ ... }` part changed, which means that the hashes changed. Ok.

PR #268, opened 2025-08-29
Fix subslice3 in glue

franziskuskiefer:
Same as we had for `subslice2`


msprotz:
Thanks -- what a drag these macros are. I'm looking forward to #259 which will make Eurydice_slice a builtin type and make these functions properly static inline.

Issue #257, opened 2025-08-07
Migrating to Monomorphised LLBC


ssyram:
Currently, Eurydice struggles with the handling of generics, partly because of the hardness and partly, even worse, due to information insufficiency from Charon. This can be partially seen from that: all the `known_failures` in `AstOfLlbc.ml` are generics. For more specific examples, the handling of associated types is hard in Eurydice; the trait resolution, if behind a very complex generic context, is also hard. Further, it is impossible for Eurydice to tell if a given type is monomorphised to a DST (dynamically sized type), given that whether an external type is DST is unknown. Not to mention computing & tracing the pointer metadata info.

Luckily, Charon is now supporting emitting fully monomorphised type with the help of Rustc itself. This helps erasing the existence of associated types and hence could also provide accurate pointer metadata info about a given type, external or not. This could also free us from handling the complex generic types during the translation. Furthermore, the monomorphised LLBC is capable of handling some features previously unable to be handled by Charon, which extends the overall capability of the tools.

Also, as the monomorphisation has been done in Charon, the translation infrastructure in Eurydice can be directly used *without modification* as from the Eurydice side, it is just translating normal objects without any generics.

But the migration might not be as obvious as it might seem to be. Thank @Lin23299 , who explored on this issue today, and found that a major problem would be to fix the Rust structure name matching, as now the item name might be changed from `func` with parameters `T1, T2` to the name of literal string `func::<T1,T2>` with no parameters. This makes the existing mechanisms like resugaring loops (required to match `iter` or so) not functioning.

Another problem associated with this is the too-long names. As now the mono mechanism of Krml is not triggered, some names get too long, as encountered in Krml before, as per discussion in #37 .

PR #275, opened 2025-09-05
Migrating to Monomorphised LLBC

Lin23299:
This PR is for #257. The current goal is to support monomorphised LLBC (add `--monomorphize` for charon in Makefile). Hopefully we can support both monomorphized and generic code. 

The first fix is to eliminate `core::option::Option` which was previouly removed by `Monomorphization.datatypes`. Now it's removed by `drop_unused_type` following the style of `drop_unused` in `krml/Inlining.ml`


Lin23299:
Now the problem is the matching of certain functions using `cremepat`. For example, the `remove_array_eq` :
```
  method! visit_expr ((n_cgs, n_binders) as env, _) e =
    match e with
    | [%cremepat {| core::array::equality::?impl::eq[#?n](#?..)<?t,?u>(?a1, ?a2) |}] ->
        let rec is_flat = function
```

was used to match
```
core.array.equality.{core::cmp::PartialEq<@Array<U,␣N>>␣for␣@Array<T,␣N>}.eq
  < int32_t > 
  < int32_t >
  [[2size_t]]
  [[core.cmp.impls.{core::cmp::PartialEq<i32>␣for␣i32}.eq]]
  uu____4@1
  uu____5@0;
```
now we want it to match the monomorphised
```
core.array.equality.{core::cmp::PartialEq::<@Array<i32,␣2usize>,␣@Array<i32,␣2usize>>}.eq.<i32,␣i32,␣2usize>
  uu____4@1
  uu____5@0;
```
Since we are not familiar with `cremepat` (I'm trying), maybe @msprotz  do you have any plan or idea to extend the pattern matching mechanism and match both mono and non-mono versions of these functions?

Lin23299:
Hi @Nadrieril @msprotz , I'm trying to fix the rewriting of `from_fn`. Now this function can be rewritten into a `for loop` using the closure function named as:
```
array.mk_incr.{core::ops::function::FnMut::<array::mk_incr::closure::<10usize>,␣(usize)>}.call_mut.<10usize>
```
While the generated function definition for it is named as:
```
array_mk_incr_{core::ops::function::FnMut::<array::mk_incr::closure::<10usize>, (usize)>}_call_mut_<10usize>
```

The names are slightly different and eurydice refuses to compile cuz it can not find the implementation of this function.
I'm confused that whether this name mismatch comes from some inconsistent naming mechanism between charon and eurydice? Or It's just my mistakes in translating the `C.trait_refs` into `K.expr list`. Any idea will be helpful :)

For reference, I'm using the test case:
```
fn mk_incr<const K: usize>() -> [u32; K] {
    core::array::from_fn(|i| i as u32)
}

fn main() {
    let a: [u32; 10] = mk_incr();
    assert_eq!(a[9], 9);
}

```

ssyram:
@protz Here's another point I want to discuss with you. First of all, I'm still not completely sure what is the third field of `ETApp`, the comments there are not very clear and I even cannot decode the English. I can understand that it collects something to be used in the later passes, but this "something" is quite messy. In `remove_slice_eq`, it refers to the internal `eq` function. While in `remove_array_from_fn`, it refers to the closure function calls. Would you mind providing a more comprehensive explanation of what they are, intuitively and as an overview, how are they computed?

Then, here comes the key question. In `expression_of_fn_ptr` for generating these third-field-content of `ETApp`, it involves a heavy computation involving the types and the generics. So there are two ways of handling them that I can conceive:
1. (Experimental & Risky) We have two completely different schemes, one just for computing the third field another for computing the ultimate type. So that, in the generic case, the first scheme simply computes useful info for the second, as original. While in the mono case, the first is completely separated and a full re-polymorphisation is performed there, while the second simply takes the mono type. As I'm not very sure what is the internal computation, the first step might be potentially risky that some info might still be missing even if we try to re-polymorphise the items.
2. (Straightforward while with Sacrifice) We keep the current mechanism and ONLY allow the built-in functions to be re-polymorphise and treat all other functions, including the impl of `eq` (used in `remove_slice_eq`), the `core::array::from_fn`, purely mono functions. This is safe and straightforward while with the explicit sacrifice that some passes like the above two are not going to work for mono codes. But as they are fully monomorphised, it should not be hard to provide manual implementation.
3. (Neutral) For just the `eq` case, there is an ugly trick of only allowing such to be re-polymorphised and the trait-refs to be collected. But I've confirmed that `from_fn` will trigger error with this partially allowance. It means we will have to manually implement `from_fn`.

ssyram:
Besides, there still seems to be the `No rule to make target 'test-array', needed by 'custom-test-array'` issue. So, I have fully tested and confirmed the problems solely for the C cases but could not yet confirm the Cpp cases.

Nadrieril:
Note: https://github.com/AeneasVerif/charon/pull/854 changes how we represent monomorphized names a tiny bit. For your purposes, you'll just have to add a few `.binder_value` and that's all.

PR #278, opened 2025-09-15
Update charon

Nadrieril:
Companion PR of https://github.com/AeneasVerif/charon/pull/796.


protz:
Thanks for the reformatting. Should we just make CI fail if something forgets to reformat?

PR #283, opened 2025-09-24
Fix CI and merge #252 

protz:


protz:
This is the same as #252 except with i) the missing fix in eurydice_glue and ii) a temporary hack (CC @Nadrieril) to pin the libcrux revision to the one that has fixes for this branch.

.github/workflows/nix.yaml
```diff
@@ -50,5 +50,6 @@ jobs:
           cd circus-green
           nix flake update charon --override-input charon github:aeneasverif/charon/$CHARON_REV
           nix flake update eurydice --override-input eurydice github:aeneasverif/eurydice/$EURYDICE_REV
+          nix flake update libcrux --override-input libcrux github:cryspen/librux/2ea4f60a8eccaf2cefbc75f6e70264d5019e22ac
```
Nadrieril: Try this: add `./update-cargo-lock.sh libcrux` before the call to `check.sh`

Nadrieril: Oh, that failed because `jq` isn't installed... Try `nix develop '.#ci' --command ./update-cargo-lock.sh libcrux` instead?

protz:
error: flake 'git+file:///run/github-runner/aeneasverif8-ci/eurydice/eurydice/circus-green' does not provide attribute 'devShells.x86_64-linux.ci', 'packages.x86_64-linux.ci', 'legacyPackages.x86_64-linux.ci' or 'ci'

protz:
ok this most recent error is beyond my nix abilities

protz:
(we're back to a missing criterion version)

.github/workflows/nix.yaml
```diff
@@ -50,5 +50,7 @@ jobs:
           cd circus-green
           nix flake update charon --override-input charon github:aeneasverif/charon/$CHARON_REV
           nix flake update eurydice --override-input eurydice github:aeneasverif/eurydice/$EURYDICE_REV
+          nix flake update libcrux --override-input libcrux github:cryspen/libcrux/2ea4f60a8eccaf2cefbc75f6e70264d5019e22ac
+          nix develop '..#ci' --command ./update-cargo-lock.sh libcrux
```
Nadrieril, line 54:
```suggestion
          nix develop --command ./update-cargo-lock.sh libcrux
```

PR #298, opened 2025-10-14
Explicit Unsized Metadata & Unified DST Handling - Split1

Lin23299:
This PR is a split for #259, it contains the main part of unified DST handling and can pass the test set of current `main` branch.

The implementations of `slice_to_array` and `slice_to_ref_array` are not included in this PR. Since the type of `Eurydice_slice` is defined as a general type in `Builtin.ml` instead of a marco type in `eurydice_glue.h`. These functions using it should be implemented using the new type in split2. In this PR we shall focus on the dst handling.

### DIFF
--- a/lib/AstOfLlbc.ml
+++ b/lib/AstOfLlbc.ml
@@ -386,6 +386,7 @@ let rec vtable_typ_of_dyn_pred (env : env) (pred : C.dyn_predicate) : K.typ =
           (* Hence we need to firstly substitute `TraitParams` with the actual types provided by the dynamic predicate's `TraitArgs` *)
           (* And then substitute `AssocTys` with the actual types provided by the dynamic predicate's assignments to these associated types *)
           (* The trait ref is guaranteed to be with empty binding values in the principal clause *)
+          (* Yet, we will need to move shift the internal DeBruijn indices with `extract_from_binder` *)
           let tref = Charon.Substitute.(extract_from_binder trait_decl_ref_substitute tref) in
           (* First step: get the `TraitArgs` *)
           let base_args =
--- a/lib/Builtin.ml
+++ b/lib/Builtin.ml
@@ -156,9 +156,8 @@ let op_128_cfgs =
 
 let get_128_op (kind, op) : K.expr = expr_of_builtin @@ Op128Map.find (kind, op) op_128_cfgs
 
-(** Get the size of the given type, corresponding to `sizeof` in C.
-    This corresponds to `NullOp::SizeOf` in Charon,
-    which is itself used in metadata field `size` in vtable. *)
+(** Get the size of the given type, corresponding to `sizeof` in C. This corresponds to
+    `NullOp::SizeOf` in Charon, which is itself used in metadata field `size` in vtable. *)
 let sizeof =
   {
     name = [ "Eurydice" ], "sizeof";
@@ -167,9 +166,9 @@ let sizeof =
     cg_args = [];
     arg_names = [];
   }
-(** Get the alignment of the given type, corresponding to `alignof` in C.
-    This corresponds to `NullOp::AlignOf` in Charon,
-    which is itself used in metadata field `align` in vtable. *)
+
+(** Get the alignment of the given type, corresponding to `alignof` in C. This corresponds to
+    `NullOp::AlignOf` in Charon, which is itself used in metadata field `align` in vtable. *)
 let alignof =
   {
     name = [ "Eurydice" ], "alignof";


include/eurydice_glue.h
```diff
@@ -88,77 +71,21 @@ typedef struct {
...
-#define Eurydice_array_to_slice(end, x, t)                                     \
-  EURYDICE_SLICE(x, 0,                                                         \
-                 end) /* x is already at an array type, no need for cast */
-#define Eurydice_array_to_subslice(_arraylen, x, r, t, _0, _1)                 \
-  EURYDICE_SLICE((t *)x, r.start, r.end)
-
-// Same as above, variant for when start and end are statically known
-#define Eurydice_array_to_subslice3(x, start, end, t_ptr)                      \
-  EURYDICE_SLICE((t_ptr)x, (start), (end))
+#define Eurydice_slice_index(s, i, t) ((s.ptr)[i])
```
protz: just to be on the safe side: `(s)`?

include/eurydice_glue.h
```diff
@@ -188,72 +115,36 @@ typedef struct {
...
+      (slice.ptr + mid), EURYDICE_CFIELD(.meta =)(slice.meta - mid)            \
+    }                                                                          \
   }
 
 #define Eurydice_slice_split_at_mut(slice, mid, element_type, ret_t)           \
   KRML_CLITERAL(ret_t) {                                                       \
-    EURYDICE_CFIELD(.fst =)                                                    \
-    KRML_CLITERAL(Eurydice_slice){EURYDICE_CFIELD(.ptr =)(slice.ptr),          \
-                                  EURYDICE_CFIELD(.len =) mid},                \
-        EURYDICE_CFIELD(.snd =) KRML_CLITERAL(Eurydice_slice) {                \
```
protz: same

lib/AstOfLlbc.ml
```diff
@@ -385,29 +359,198 @@ let lookup_field env typ_id field_id =
...
+                      && assn.type_name = name
+                    then
+                      Some assn.ty
+                    else
+                      None
+                  in
+                  match List.find_map finder assoc_ty_assns with
+                  | Some ty -> ty
+                  | None ->
+                      fail "Could not find associated type assignment for associated type %s" name)
```
protz: this is not doing what you want, you need to wrap the match in begin/end

lib/AstOfLlbc.ml
```diff
@@ -385,29 +359,198 @@ let lookup_field env typ_id field_id =
...
+          (* The trait ref is guaranteed to be with empty binding values in the principal clause *)
+          let tref = Charon.Substitute.trait_decl_ref_substitute out_of_binder_subst tref in
+          (* First step: get the `TraitArgs` *)
+          let base_args =
+            let generics = tref.generics in
+            { generics with types = List.tl generics.types }
+          in
+          (* Second step: find from the assignments of assoc tys *)
+          let assoc_tys =
+            let rec drop n l =
```
protz: use `Krml.KList.split_at`? generally, if you have little list helpers like this, it's more readable if you move them out of the main logic

lib/AstOfLlbc.ml
```diff
@@ -385,29 +359,198 @@ let lookup_field env typ_id field_id =
...
+         monomorphized LLBC."
+  (* The metadata of a &dyn Trait object is a pointer to its vtable *)
+  | C.TDynTrait pred -> Some (K.TBuf (vtable_typ_of_dyn_pred env pred, false))
+  | C.TLiteral _ | C.TRef (_, _, _) | C.TRawPtr (_, _) | C.TFnPtr _ | C.TFnDef _ -> None
+  (* The metadata must not have ptr metadata as they must be Sized. *)
+  | C.TPtrMetadata _ -> None
+  | C.TNever | C.TError _ -> failwith "Error types to fetch metadata"
+
+(* Get the pointer type of the original Charon type. *)
+and ptr_typ_of_ty (env : env) (ty : Charon.Types.ty) : K.typ =
```
protz: suggest different comment: "Translate Charon type &T as a krml type -- this handles special cases where address-taking in Rust creates a DST, which we represent as instances of the `dst_ref` type. There are many such cases: slices, &str, etc."

lib/AstOfLlbc.ml
```diff
@@ -1618,19 +1743,48 @@ let is_str env var_id =
...
+            with_type
+              (Builtin.mk_dst_ref t metadata.typ)
+              (EFlat [ Some "ptr", ptr; Some "meta", metadata ]))
+      | _ ->
+          K.(
+            with_type
+              (Builtin.mk_dst_ref e.typ metadata.typ)
+              (EFlat [ Some "ptr", addrof e; Some "meta", metadata ])))
+
+(* Currently unused when metadata of CastUnsize is given
```
protz: remove this

lib/AstOfLlbc.ml
```diff
@@ -1773,6 +1879,20 @@ let expression_of_rvalue (env : env) (p : C.rvalue) expected_ty : K.expr =
       else
         failwith
           ("unknown cast: `" ^ Charon.PrintExpressions.cast_kind_to_string env.format_env ck ^ "`")
+  (* | UnaryOp (PtrMetadata, e) ->
```
protz: let's remove

lib/AstOfLlbc.ml
```diff
@@ -1723,26 +1831,24 @@ let expression_of_rvalue (env : env) (p : C.rvalue) expected_ty : K.expr =
...
       let t_from = typ_of_ty env ty_from and t_to = typ_of_ty env ty_to in
       let e = expression_of_operand env e in
       begin
-        match t_from, is_dst env t_to with
-        | ( TBuf (TApp (lid1, [ K.TCgApp (K.TApp (lid_arr, [ u1 ]), cg) ]), _),
-            Some (lid2, TApp (slice_hd, [ u2 ]), t_u) )
-          when lid1 = lid2 && u1 = u2 && slice_hd = Builtin.derefed_slice && lid_arr = Builtin.arr
-          ->
+        match meta, t_from, is_dst_ref t_to with
+        | MetaLength cg, TBuf (TApp (lid1, _), _), Some (lid2, t_u) when lid1 = lid2 ->
```
protz: let's confirm here why the check that this is an array is gone -- there was `lid_arr = Builtin.arr` in the guard above, but it's now gone

lib/Builtin.ml
```diff
@@ -524,6 +516,193 @@ let layout =
     arg_names = [];
   }
 
+(* Implementations, defined as DFunction using abstract syntax, mainly for generic slice type *)
+
+(* Helpers for common use *)
+
+let ws = K.with_type (TInt SizeT)
+
+(* s[i] where s:t[] *)
```
protz: if these are helpers, maybe call them mk_...?

lib/Builtin.ml
```diff
@@ -524,6 +516,193 @@ let layout =
...
+let index t s i =
+  let slice_index = expr_of_builtin_t slice_index [ t ] in
+  K.with_type t (K.EApp (slice_index, [ s; i ]))
+
+(* left value of s[i] *)
```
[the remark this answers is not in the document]
protz: not used elsewhere, delete

lib/Builtin.ml
protz: let's remove if not used

lib/Builtin.ml
```diff
@@ -524,6 +516,193 @@ let layout =
...
+let lhs t s i =
+  K.(
+    with_type t
+      (EBufRead (with_type (TBuf (t, false)) (EAddrOf (index t s i)), Krml.Helpers.zero_usize)))
+
+(* a->data where a: Arr<t,n>*  *)
+let data_of_arrref (arrref : K.expr) t n_cgid =
```
protz: comment: "to be used before eurydice monomorphization"

lib/Builtin.ml
```diff
@@ -524,6 +516,193 @@ let layout =
...
+    (* args *)
+    let n = ws (EBound 1) in
+    let arrref = with_type arrref_t (EBound 0) in
+    (* { .ptr = a->data , .meta = N }*)
+    let data = data_of_arrref arrref element_t 0 in
+    with_type ret_t (EFlat [ Some "ptr", data; Some "meta", n ])
+  in
+  DFunction (None, [ Private ], 1, 1, ret_t, lid, binders, expr)
+
+let array_to_subslice_func =
```
protz: likewise, and all other functions below

lib/PreCleanup.ml
```diff
@@ -76,9 +76,74 @@ let remove_array_eq =
       super#visit_DFunction (n_cgs, 0) cc flags n_cgs n t lid bs e
   end
 
+(** Comes from [drop_unused] in Inlining.ml, we use it to remove the builtin function defined using
+    abstract syntax when they are not used. Otherwise they may use some undefined types and fail the
+    check *)
```
protz: just say here that it's because of range

lib/AstOfLlbc.ml
Lin23299: It is *intentional* because we now support not only from `T<[U;N]>` to `T<[U]>` but also from `T<S1>` to `T<S2>` for user-defined DST types. So this guard should be removed.

The test case for this is `test/more_dst.rs` :
```
struct S<U: ?Sized> {
    foo: u32,
    data: U,
}

type T1 = S<[u32]>;
type TS1 = S<[u32; 4]>;

type T2 = S<T1>;
type TS2 = S<TS1>;

type T3 = S<T2>;
type TS3 = S<TS2>;

fn main() {
    let x1 = TS1 { foo: 0, data: [0; 4]};
    let y1 : &T1 = &x1;
    assert_eq!(y1.data[1],0);

    let x2 = TS2 { foo: 0, data: x1};
    let y2 : &T2 = &x2;
    assert_eq!(y2.data.data[2],0);

    let x3 = TS3 { foo: 0, data: x2};
    let y3 : &T3 = &x3;
    assert_eq!(y3.data.data.data[3],0);
}
```
We can now handle such casting for multi-level user-defined DSTs.

protz:
Ok, for the build failure, please run `git rm -rf out/ && make test`, then commit the result -- some files (like Eurydice.h) are now gone from the code generation.

For instance: `git add out && git commit -am "Refresh test output"`

lib/AstOfLlbc.ml
Lin23299: It's already updated [here] (https://github.com/ssyram/eurydice/blob/33ecfe8d7b0fecd9e4d76c7bcfeeac8de70f9603/lib/AstOfLlbc.ml#L1831-L1835) as follows
```
  | UnaryOp (Cast (CastUnsize (ty_from, ty_to, meta) as ck), e) ->
      (* DSTs: we support going from &T<S1> to &T<S2> where S1 is sized,  S2 is
         unsized and &T<S2> becomes a fat pointer. The base case is from &T<[U;N]>
         to T<[U]>. See test/more_dst.rs for user-defined DST case. We build this 
         coercion by hand, and slightly violate C's strict aliasing rules. *)
```

protz:
Ok the error is related to slice_to_array2

Lin23299:
> slice_to_array2

Then we need the split2 of this PR :)

lib/AstOfLlbc.ml
```diff
@@ -385,29 +359,198 @@ let lookup_field env typ_id field_id =
...
+    [dyn Trait<TraitArgs, AssocTy=T, ...> + AutoTraits + 'a]. Namely, it has the arguments to the
+    unique non-auto trait and the assignments to each associated types as the **principal trait**
+    [Trait<...>], and then some additional auto-traits having no methods and finally a region
+    constraint. Here we care only about the principal trait. *)
+let rec vtable_typ_of_dyn_pred (env : env) (pred : C.dyn_predicate) : K.typ =
+  let binder_params = pred.binder.binder_params in
+  match binder_params.trait_clauses with
+  | [] -> failwith "DynTrait has empty clause! Please report this to Charon."
+  (* As guaranteed by Rustc and hence Charon, the first clause must be the principal trait *)
+  | principal_clause :: _ -> (
```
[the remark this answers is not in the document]
Lin23299: It IS required by the format-apply

lib/AstOfLlbc.ml
Lin23299: It's now wraped in begin/end.

lib/AstOfLlbc.ml
Lin23299: Replaced using `Krml.KList.split`.

protz:
@Lin23299 CI passes here.

TODO to merge this:
1. Please describe briefly what remains on this branch (sizeof and alignof + better way of doing substitutions).
2. Please make sure there's a test for sizeof and alignof

protz:
Great thanks!

PR #299, opened 2025-10-14
Supporting [slice_to_array] under unified DST handling - Split2

Lin23299:
This PR reimplements `slice_to_array` and `slice_to_ref_array` under the unified DST handling.

The ideas can be summarized as follows:

1. The argument type is now a builtin generic type (`dst_ref`) instead of the `Eurydice_slice` defined as C macro. So we should use abstract syntax to defined these functions.

2. The return type is now a struct for array instead of a pointer. So we need to first allocate a struct and then `memcpy` the content of the slice into the `data` field of the array struct.

3. To implement them, we need the type of the elements of slice/array. So I use `expand_slice_to_array` in `PreCleanup.ml` to get the element type and rewrite these function calls instead of providing function definitions for them like other builtin funcitons. Maybe it is also possible to extract the element type from the argument type of slice, but I don't know for now.

4. For `slice_to_ref_array`, the arguments includes a length and this function returns `Err` is it is bigger than the length of the incoming slice (See `Eurydice_slice_to_ref_array2` in `eurydice_glue.h`). I reused this C macro to implement this behavior instead of rewriting it using abstract syntax.

### DIFF
--- a/out/test-libcrux/internal/libcrux_core.h
+++ b/out/test-libcrux/internal/libcrux_core.h
@@ -17,12 +17,6 @@ extern "C" {
 
 #include "../libcrux_core.h"
 
-static inline uint32_t core_num__u32__count_ones(uint32_t x0);
-
-#define CORE_NUM__U32__MAX (~0U)
-
-#define CORE_NUM__U32__BITS (core_num__u32__count_ones(CORE_NUM__U32__MAX))
-
 static inline uint64_t core_num__u64__from_le_bytes(Eurydice_arr_c4 x0);
 
 static inline uint64_t core_num__u64__rotate_left(uint64_t x0, uint32_t x1);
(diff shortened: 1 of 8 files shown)


protz:
My intuition is that because this happens pre-monomorphization, you may have a const-generic as the array length. This is fine, and there should be helpers to deal with that.

protz:
cg_of_expr in lib/DeBruijn.ml should have comments, but I'm surprised we never had to convert a const-generic argument to an expression before...

lib/PreCleanup.ml
```diff
@@ -76,6 +76,91 @@ let remove_array_eq =
       super#visit_DFunction (n_cgs, 0) cc flags n_cgs n t lid bs e
   end
 
+let expression_of_cg (cg : cg) =
+  match cg with
+  | CgConst n -> Krml.Helpers.mk_sizet (int_of_string (snd n))
+  | CgVar _ -> failwith "non-const length in the Arr<T,C>"
+
+let expand_slice_to_array =
```
protz: Ok so I'm now reviewing this. I think what would help is if you showed the AST before and the AST after, in concrete syntax. Also, what is the original type of slice_to_array?

protz:
please merge branch `protz/fix-binders` into this branch to see if this fixes CI

protz:
Ok so #301 fixes the const generic issue, and now we are down to issues in kyber:

No implementation for this:

```
ml-kem>  3178 |       repeat_expression[0U] = core_array___Array_T__N___as_slice(
```

Furthermore:

```
 ml-kem> /build/libcrux-src/libcrux-ml-kem/c/internal/../intrinsics/libcrux_intrinsics_avx2.h:98:5: error: unknown type name 'Eurydice_slice'; did you mean 'Eurydice_vec'?
```

should we have a #define in the glue to facilitate backwards-compatibility?

The rest of the errors will need some intervention from the libcrux folks, unless you want to fix issues yourself, in which case I can provide pointers. Thanks.

protz:
Ok from now on it's just renamings of types in libcrux.

Lin23299:
I've made a set of renaming and defined `&slice[u8]` and `&slice[i16]` in `eurydice_glue.h` in order to define the builtin functions in `test/libcrux`, basically using the same method with defining `array[u8;2]` for eurydice builtins, i.e. droping these definitions in `Cleanup3.remove_builtin_decls`. Now the `test-libcrux` success in my PC but there is still something wrong in the online CI that I cannot understand. So @protz could you take a look?

protz:
Wonderful! I think there's not much left to fix. I'll take a look

protz:
Congratulations @Lin23299!

PR #306, opened 2025-10-21
Move libcrux CI into Eurydice test suite

### LINKED ISSUE #290: CI strategy to allow fixing multiple things in one go
After many iterations of the plan, the current goal is to treat libcrux as part of the regression suite for Eurydice, and specifically, to have build descriptions and/or toolchain invocations for libcrux within the tree of Eurydice. Operationally, this means that `make test` will depend on `make libcrux-test`, which will:
  - warn if the version of libcrux found via LIBCRUX_HOME is different from the one found in the flake.lock
  - run the Charon invocation to get .llbc files out of the libcrux tree
  - run the Eurydice invocation to get C code
  - use CMakeLists and C++ tests *in the Eurydice source tree* to build the resulting C code and run tests for it

CI will take care of fetching the relevant libcrux version, and exporting LIBCRUX_HOME.

One might say that circus-green has less value in this scenario; that's correct. But I still think there's value in catching unintentional breakage, such as:
- Eurydice breaks libcrux by accident, because libcrux exercises some newer feature in its latest `main`, which Eurydice is not yet using for its regression suite

The rationale for this is as follows:
- some PRs require changes on the libcrux side and these are very hard to coordinate (see e.g. #280 -- I need to fix both the way libcrux invokes charon, and something on the eurydice side; having both in the same place would allow me to move forward in one go)
- this allows Eurydice to maintain a greater test coverage than what libcrux currently exercises -- right now, libcrux has committed to maintaining only the header-only versions of mlkem and mldsa, which is a good test case, but does not necessarily cover everything
- generally, this allows a simpler contract where eurydice treats libcrux as a first-tier testcase, meaning that there is *always* a way to extract libcrux; however, libcrux itself might be using an older version of the toolchain

The drawback is having to maintain some build and test descriptions, but that's fine.

The final issue is that libcrux currently only extracts (and tests only run, insofar as I understand) if you're on x64. That is, the output changes on arm64. That's the one thing that bothers me and I'd love to hear some thoughts.

CC @franziskuskiefer @kek @Nadrieril

protz:
This has the following effects.
- Build description (config.yaml) and tests (in C++) move to the Eurydice side
- libcrux output is now under version control to assess the impact of Eurydice changes
- this means that should the output change (intentionally), then we can fix the tests too instead of coordinating updates between several repositories.

The intent is that the latest, "correct" way to build libcrux is showcased in this repository. When libcrux wishes to update its toolchain, it should look here to see how to update.

This fixes #290


protz:
CC @franziskuskiefer and @keks

protz:
Thanks @Nadrieril !

PR #308, opened 2025-10-27
Vtable Instance Support

ssyram:
Continuing from #259, we propose another PR to support vtable instances, specifically generating concrete vtable pointers. Currently, Eurydice only supports generating vtable structs but not their instances.

We added a test by renaming `dyn_trait_struct_type` to `dyn_trait` and included a test case with `use_trait(&100)`, which involves casting `&i32` to `&dyn Trait`.

Charon currently lacks proper implementation for the `drop` field in vtable instances and implicit implementations. The latter is unavoidable since every vtable requires `T: MetaSized`, which relies on an implicit implementation. To address this, we introduced a comprehensive placeholder for `Opaque` values via a new built-in function `Eurydice::opaque<T>(reason: char*) : T`. This should resolve currently unhandled cases.

For the `drop` field, we temporarily implemented a Charon version that generates a placeholder value. Therefore, this PR can only be merged after `drop` is properly implemented, and it would be preferable to wait until implicit implementations are complete. The temporary version we used is https://github.com/ssyram/charon/tree/tmp-empty-drop.

However, two issues remain even after Charon is fully prepared:

## Vtable Instance Initialisation

The primary concern involves the generated code for vtable instances. Currently, we produce:

```C
static size_t _vtable__local_0 = Eurydice_sizeof(int32_t, size_t);
static size_t _vtable__local_1 = Eurydice_alignof(int32_t, size_t);

static dyn_trait_Trait__vtable_ _vtable__local_2 = {
    .size = _vtable__local_0,
    .align = _vtable__local_1,
    .drop = dyn_trait__vtable_drop_shim__92,
    .method_method = dyn_trait_method__vtable_method__92,
    .super_trait_0 = Eurydice_opaque(
        "missing supertrait vtable",
        core_marker_MetaSized__vtable_*,
        core_marker_MetaSized__vtable_*
    )
};

const dyn_trait_Trait__vtable_ dyn_trait__dyn_trait__Trait_for_i32___vtable_ = _vtable__local_2;
```

This code is incompatible with C compilers because the initializations of `_vtable__local_2` and `dyn_trait__dyn_trait__Trait_for_i32___vtable_` involve non-constant expressions. The only viable solution is to aggressively inline local variables, particularly within global variables.

## Opaque handling

Also, the implementation of `opaque` now also raises compiler complaints: `error: indirection of non-volatile null pointer will be deleted`. I'm not sure if this is a problem -- I guess it should raise compilation complain as opaque itself is not acceptable for code compilation.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -57,7 +57,7 @@ clean-and-test:
 .PRECIOUS: %.llbc
 %.llbc: %.rs .charon_version
 	# --mir elaborated --add-drop-bounds 
-	$(CHARON) rustc --preset=eurydice --dest-file "$@" $(CHARON_EXTRA) -- $<
+	$(CHARON) rustc --preset=eurydice --dest-file "$@" $(CHARON_EXTRA) -- -Aunused $<
 
 out/test-%/main.c: test/main.c
 	mkdir -p out/test-$*
@@ -155,11 +155,12 @@ test/libcrux.llbc:
 	RUSTFLAGS="-Cdebug-assertions=no --cfg eurydice" $(CHARON) cargo --preset eurydice \
 	  --include 'libcrux_sha3' \
 	  --include 'libcrux_secrets' \
+	  --rustc-arg='-Aunused' \
 	  --start-from libcrux_ml_kem --start-from libcrux_sha3 \
 	  --include 'core::num::*::BITS' --include 'core::num::*::MAX' \
 	  --dest-file $$PWD/$@ -- \
 	  --manifest-path $(LIBCRUX_HOME)/libcrux-ml-kem/Cargo.toml \
-	  --target=x86_64-apple-darwin 
+	  --target=x86_64-apple-darwin
 	@# Commit the `Cargo.lock` so that the nix CI can use it
 	cp $(LIBCRUX_HOME)/Cargo.lock libcrux-Cargo.lock
 
--- a/README.md
+++ b/README.md
@@ -38,8 +38,11 @@ the [Zulip](https://aeneas-verif.zulipchat.com/).
 # Install
 
 We recommend using Nix to easily ensure you are running the right versions of the tools and
-libraries. Our CI uses the `flake.lock` and `flake.nix` files, meaning that you are always
-guaranteed a successful build if you use Nix.
+libraries. With nix, you can run:
+```bash
+$ nix run 'github:aeneasverif/eurydice#charon' -- [CHARON_OPTIONS]
+$ nix run 'github:aeneasverif/eurydice' -- [EURYDICE_OPTIONS] <llbc_file>
+```
 
 Alternatively, you can do a local setup as follows.
 
--- a/flake.lock
+++ b/flake.lock
@@ -26,11 +26,11 @@
         "rust-overlay": "rust-overlay"
       },
       "locked": {
-        "lastModified": 1764614553,
-        "narHash": "sha256-OFY2wVBn8UB0JPDUN7AsCpPipOSrS9T7NBD0MaCaX9Q=",
+        "lastModified": 1767093296,
+        "narHash": "sha256-xVO/cOWzKJFcht6tkqXA9aErR02Xdj4XxKDgt5Gn54Q=",
         "owner": "aeneasverif",
         "repo": "charon",
-        "rev": "377317d6b25702c46ffff072fa00a3e32095e46f",
+        "rev": "e587fcdad522736b216c23563cb6a7715450844d",
         "type": "github"
       },
       "original": {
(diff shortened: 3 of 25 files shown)


ssyram:
Thanks for the discussion! We agree with the first point. As for the second, the `Eurydice::opaque<T>` is a generic function, so Eurydice automatically drops it after its mono phase and requires us to provide the impl?

protz:
You could implement it with a macro that uses `KRML_HOST_EABORT` to abort cleanly and return a value of type T?

ssyram:
Hi @protz we have added the generics pushing there, so that `AstOfLlbc` now can safely translate the stuff, but the problem then becomes that after the monomorphization of the functions is not finished properly --- 
```OCaml
  (**)
    auto_generated function
  core_fmt_fmt__vtable_method__eb_7c
  <cg: 0><0>(
    mutable
    _(Mark.Present,1, ):
    Eurydice_dst_ref
      Eurydice_c_void_t
      core_fmt_Debug_{vtable}*,
    mutable
    _(Mark.Present,1, ):  core_fmt_Formatter*
  ):
  core_result_Result
  () core_fmt_Error
  {
    core.fmt.{core::fmt::Debug␣for␣&0␣(T)}.fmt < 0 > □
      ((_@1).ptr<:   0**)
      _@0
  }
```

This is quite strange as it should be a function monomorphization instead of global monomorphization. But still, after the mono, there is still a call to `TBound 0` in the function body. Could you please have a look?

FYI, we are testing against the example:
```Rust
fn use_debug(_d : &dyn std::fmt::Debug) {}
fn main() { use_debug(&&1) }
```

With the Charon branch in https://github.com/ssyram/charon/tree/tmp-empty-drop -- this branch now generates an empty stub for the `drop` fields.

ssyram:
OK, we then further refined the problem, and added the generics to globals (see the latest two commits). But then Krml complains that:
```"lib/krml/Checker.ml", line 126, characters 10-16: Assertion failed```
We found that it is to assert that globals all should have *empty* generic types (also, globals do not have CGs):
```OCaml
| DGlobal (_, lid, n, t, _) ->
          assert (n = 0);
          { env with globals = M.add lid t env.globals }
```

This is in Krml, so we are not sure if it is a fundamental problem. Could you please have a look?

FYI, we are testing against the example:
```Rust
fn use_debug(_d : &dyn std::fmt::Debug) {}
fn main() { use_debug(&&1) }
```

With the Charon branch in https://github.com/ssyram/charon/tree/tmp-empty-drop -- this branch now generates an empty stub for the `drop` fields.

protz:
I will have a look, thank you.

Can globals have const generics in Rust?

ssyram:
Thanks for this! We've tried the method, but we then found that it is more complicated than conceived that we will need to collect the `fn_ptrs` of traits as in the function pointers. As the generics for globals is just for this case, I'm personally considering ignoring / bypassing this problem by simply emitting an `EAbort` as the value and wait for the mono migration (#257) to complete.

Out of curiosity, I'm not sure that for example, in [line 1887 of this PR](https://github.com/AeneasVerif/eurydice/pull/308/files#diff-8a0b9592c1b552914e953891fc9e0d912bfd83270c97e74e8579e6dc574159d8R1887), is this substitution a correct way or we should recover the `TPoly` type and use `TApp` instead?

protz:
Is the vtable type a parameterized data type T<'a> and instances have type T<U>? If so, yes, ETApp is necessary because it will trigger monomorphization

Lin23299:
I've tried to update this PR and it works. There is only a little remaining problem that the compiled `dyn_trait.h` contains duplicated `typedef`s :
```
typedef struct test_Trait__vtable__s test_Trait__vtable_;

typedef struct test_Trait__vtable__s test_Trait__vtable_;

typedef struct test_Trait__vtable__s test_Trait__vtable_;

typedef struct test_Trait__vtable__s
{
  size_t size;
  size_t align;
  void (*drop)(Eurydice_dst_ref_mut_b8 x0);
  void (*method_method)(Eurydice_dst_ref_mut_b8 x0);
  const core_marker_MetaSized__vtable_ *super_trait_0;
}
test_Trait__vtable_;

```
The LLBC file looks normal, and in the log it looks like:
```
  type  test_Trait_{vtable}  =<cg: 0> forward struct

  type  test_Trait_{vtable}  =<cg: 0> forward struct

  type  test_Trait_{vtable}  =<cg: 0> forward struct

  type  test_Trait_{vtable}  =<cg: 0>
    flat
    {
     ...
    }

```
 I will try to locate this later, @protz  @ssyram  Do you have any clues that any code in the recent commits may cause this issue?

Issue #312, opened 2025-10-28
array_to_subslice3 was not restored following the DST PR


protz:
7d4d0fd725397ce110e47322cde4f53599e999c7 removed the definition of array_to_subslice3 and slice_subslice3

The goal of this phase was to replace occurrences of array_to_subslice (and slice_subslice3) where the range start and end are constant values, with a more optimized macro that does not take so many arguments.

Concretely, the removal of array_to_subslice3 generates a diff like this:

```diff
-  Eurydice_slice_copy(Eurydice_array_to_subslice3(uu____0,
-      (size_t)0U,
-      Eurydice_slice_len(slice, uint8_t),
-      uint8_t *),
+  Eurydice_slice_copy(array_to_subslice_3621(uu____0,
+      (
+        KRML_CLITERAL(core_ops_range_Range_08){
+          .start = (size_t)0U,
+          .end = Eurydice_slice_len(slice, uint8_t)
+        }
+      )),
```

The new code is quite a bit more verbose.

Can we restore this code quality optimization? @Lin23299 thanks!

Issue #311, opened 2025-10-28
Generic Global Variables


ssyram:
We found that currently Eurydice does not support generic global variables. Originally, user-defined generic global variable is rare, but now, as we generate for each `impl Trait for T` a vtable instance for `Trait`, which is in turn a global variable, it becomes a common issue.

For example, if we test the following code like that in AeneasVerif/charon#873 :
```Rust
fn use_debug(_d: &dyn core::fmt::Debug) { }
fn main() { use_debug(&&1); }
```

It involves the `impl <T: Debug> Debug for &T`, and hence involves the generic global variable. Will there be a possible quick fix for this now? Like, can we share the logic of generic handling from functions to also globals?

> We understand that this is a feature going to be deprecated in the future by mono migration, but it would be convenient at least for now to have this issue resolved.

ssyram:
Oh, according to @Lin23299 , he found that the `Not_found` problem comes from that the generic parameters are not properly pushed in the binders so that when translating with `typ_of_ty`, it says it cannot find the type parameter `T`.

He has also identified that in the `AstOfLlbc` file, the global translation does not push the generic binders as it is for functions. So we were curious about whether this is a design intention or just a bug that we can quickly fix.

cjpopova:
Adding another example - as hinted before, this issue affects derived debug print formatting, for example:
```rust
#[derive(Debug)]
struct a {
  b : u8
}

pub fn print_struct() {
  let my_a = a { b : 1u8 };
  println!("{:?}", my_a.b);
}
```
->
```
ERROR translating derive_debug::{core::fmt::Debug<derive_debug::a>}::fmt
```
Debug formatting is quite prevalent, especially in some compiler fuzzers like [Rustsmith](https://github.com/rustsmith/rustsmith) and [Rustlantis](https://github.com/cbeuw/rustlantis) (although it is not strictly necessary in the latter).

protz:
@cjpopova the fix is described in a comment, do you think you can submit a PR with a change?

PR #313, opened 2025-10-29
Restored array_to_subslice3 & code quality optimizations following the DST PR

Lin23299:
According to #312 , I first tried to restore this optimization using `array_to_subslice3`. I think the result is ok but what's happening here feels strange for me.

At this phase (`PreCleanup3`), we are rewriting `array_to_subslice` using `Range` into a `array_to_subslice3` using 2 `size_t`s , but the difference here is that we have the monomorphized definition of `Eurydice_array_to_subslice_XX`. We have to remove it after this optimization. So basically we define the function `array_to_subslice` using abstract syntax with `Dst_Ref`, and replace it later with a C macro by passing the monomorphized return type `Dst_Ref_XX`.

Back to the quality issue, it seems to me that the point is to get rid of the `Range` type. If so, maybe a better solution is to rewrite them earlier before the monomorphization, therefore we can get a monomorphized func def of `array_to_subslice3_XX` using `size_t`s instead of  C macro definition?

For the general optimization point, I agree that we could have many improvements. Following the discussion above, I think the question is to choose from:
1. builtin functions + their monomorphized implementations in the generated C file (the current version)
2. builtin functions + C macro definitions (the version before DST PR)
3. directly rewriting these functions using abstract syntax (e.g. slice_index(s,n,t) -> s.ptr[n])

I'm not sure which one is better, my previous plan is based on 1. Actually the current version is a mixup of 1 and 2 cuz I only wrote the abstract syntax for the functions could not be defined using macro. The `array_to_subslice3` issue implies that it is possible to switch from 1 back to 2 for these functions. @protz 

Edit: I just found that the current macro definition causes stack smashing in `test-libcrux` and I don't know why. This is a reason I personally prefer 1 than 2 because C macro is more error-prone.

### DIFF
--- a/lib/AstOfLlbc.ml
+++ b/lib/AstOfLlbc.ml
@@ -2753,12 +2753,37 @@ let file_of_crate (crate : Charon.LlbcAst.crate) : Krml.Ast.file =
   let declarations = flatten_declarations declarations in
   seen := 0;
   total := List.length declarations;
-  let get_nth_function id = C.FunDeclId.Map.find id fun_decls in
-  let get_nth_type id = C.TypeDeclId.Map.find id type_decls in
-  let get_nth_global id = C.GlobalDeclId.Map.find id global_decls in
-  let get_nth_trait_impl id = C.TraitImplId.Map.find id trait_impls in
-  let get_nth_trait_decl id = C.TraitDeclId.Map.find id trait_decls in
   let format_env = Charon.PrintLlbcAst.Crate.crate_to_fmt_env crate in
+  let get_nth_function id =
+    match C.FunDeclId.Map.find_opt id fun_decls with
+    | Some f -> f
+    | None ->
+        fail "Function id not found: %s"
+          (Charon.PrintExpressions.fun_decl_id_to_string format_env id)
+  in
+  let get_nth_type id =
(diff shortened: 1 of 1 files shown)


Lin23299:
> You are correct. Having these as a late-phase cleanup that is untyped was necessary only because the definition of slice was 
> I think 2. should be definitely ruled out, for the reasons that you mention.
> 
> I think we should do a mix of 1. and 3. I think it's mostly going to boil down to a cosmetic / taste factor, as to what we think is most readable in the C code:
> 
> * `slice_len(e)` ~~> `e.meta` is definitely a case of 3., but can be done as an earlier phase that is well-typed; same deal with slice_index, etc.
> * `array_to_subslice` with a constant Range argument can become a call to `array_to_subslice_range` (let's pick a more readable name than array_to_subslice_3!), which perhaps is more of a case of 1... or do you think that `(Eurydice_slice_XX)({ .ptr = arr + start, .len = end - start })` is more readable?
> 
> An additional complication is that `fn<const K>(x: [T; N]) { x.to_subslice(0, K) }` will be have constant arguments only _after_ monomorphization, so perhaps we will miss some opportunities to simplify calls to `array_to_sublice`
> 

Totally agreed. I'll try to implement these optimizations after helping @ssyram  to fix the issues on #311

Issue #310, opened 2025-10-28
Potential Bug in Generic Handling


ssyram:
> Just a recording of a potential bug, causing no error for now but might be somehow harmful in the future

Currently, in the (longer version of) function `expression_of_place`, when one tries to translate `PlaceGlobal`, the code is:
```OCaml
      let global = env.get_nth_global id in
      K.with_type (typ_of_ty env global.ty) (K.EQualified (lid_of_name env global.item_meta.name))
```

This is potentially buggy -- we are fetching the original **declaration** type of the `global` instead of the **usage** type -- notably, there might be generics, so the declaration may contain some variables referring to its **declaration generic params**, but for the **usage**, we actually have the generic args, hence, we should substitute the generics inside with the args before calling the `typ_of_ty`.

There is an easy fix for this, of course, and it has actually been implemented in #308 in its vtable instance translation, but I'm not quite sure if there is still likewise issues in other places of the code. Hence would like to raise an issue on the overall situation.

ssyram:
OK, so if there is no potential use case, I think we can ignore this and simply wait for the mono migration? @Nadrieril seems to have a strong view of eliminating generic support (not the infrastructure) in Eurydice, i.e., Eurydice must be invoked with mono LLBC in this future.

protz:
It might be good to take the fix out of #308 so that I can review that fix independently. It'll make #308 smaller to review.

ssyram:
OK, will have a try.

PR #314, opened 2025-10-30
simple fix of types of globals

ssyram:
As mentioned in #310 , here's a fix to the problem mentioned there. But instead of the more complex substitution as in #308 , we simply fix it by using the tagged type of the place.

Notably, in #308 , we will need to fetch the vtable instance pointer, and substitute the generics with the generics args from the reference to generate the correct type ourselves. But in `PlaceGlobal`, we have available type tagged by Charon.

Further, please note that the tagged type may not yield exactly the same result as the substitution -- because of the difference in lifetime parameters.

As mentioned in #310 , there might be more hidden bugs like this.


protz:
Thanks!

PR #317, opened 2025-11-03
Remove phase that is now un-necessary after #252

protz:
FYI @ssyram and @Lin23299


ssyram:
Thanks! This is good! I guess `pass_by_ref` is also no longer functioning after the array PR?

protz:
Yes did we remove it? If not, probably worth another PR

PR #341, opened 2025-11-14
Grab a more recent karamel and fix #338

### LINKED ISSUE #338: Bug: Assigning to `false` with `--mir elaborated`
The following codes:
```Rust
enum Struct {
  A(String),
  B(Box<i32>),
  C
}
enum ThreeWays {
  Middle(i32),
  Left(String),
  Right(Box<f64>),
}
fn use_struct(s: ThreeWays) -> Struct {
  match s {
    ThreeWays::Middle(i) => Struct::B(Box::new(i)),
    ThreeWays::Left(st) => Struct::A(st),
    _ => Struct::C,
  }
}
fn main() { }
```

When running  Charon with `--mir elaborated`, e.g.,
> `./charon/bin/charon rustc --preset=eurydice --mir elaborated --print-llbc -- test.rs > output.txt`

Eurydice produces the following error:
```
Cannot re-check test.use_struct as valid Low* and will not extract it.  If test.use_struct is not meant to be extracted, consider marking it as Ghost, noextract, or using a bundle. If it is meant to be extracted, use -dast for further debugging.

Warning 4: in the sequence statement at index 0, in top-level declaration test.use_struct, in file test: Malformed input:
EAssign wants a lhs that's a mutable, local variable, or a path to a mutable field; got false instead
```

This is because there is a strange assignment in the AST during cleanup:
```Rust
false := true
```

protz:


Lin23299:
The point is the type `String` which requires drop function on the heap, and the boolean assignments was for these drops. Replacing it by `u8` makes the llbc result do not contain the ` a:= false; a:= true` anymore so the issue cannot be reproduced I think.

Using `Vec<u8>` reproduces the issue on my PC and causes the following change in output C:
```
diff --git a/out/test-lvalue/lvalue.c b/out/test-lvalue/lvalue.c
index b3fbb96..445aa00 100644
--- a/out/test-lvalue/lvalue.c
+++ b/out/test-lvalue/lvalue.c
@@ -14,11 +14,13 @@ void lvalue_main(void)
 
 lvalue_Struct lvalue_use_struct(lvalue_ThreeWays s)
 {
-  lvalue_Struct uu____0;
+  bool uu____0 = false;
+  uu____0 = true;
+  lvalue_Struct uu____1;
   if (s.tag == lvalue_Middle)
   {
     int32_t i = s.val.case_Middle;
-    uu____0 =
+    uu____1 =
       (
         KRML_CLITERAL(lvalue_Struct){
           .tag = lvalue_B,
```

and the C code 
```c
bool uu____0 = false;
uu____0 = true
```
is rejected by compiler as [-Werror=unused-but-set-variable]. It seems that the current cleanups in eurydice and karamel does not handle such deadcode?

protz:
Ok I like this change. You should be able to add that to the code, and then do:

```
test-lvalue: CFLAGS += -Wnounused-but-set-variable
```

in the Makefile to make this just a warning, not an error

protz:
So to be clear, 1. edit the test to have Vec<u8> and then 2. fix the Makefile to make "set but unused" not an error. 

Then I'll merge this. Thanks!

Lin23299:
> So to be clear, 1. edit the test to have Vec and then 2. fix the Makefile to make "set but unused" not an error.

I do not have the permit to push on your branch, so I've pushed it on the fork https://github.com/ssyram/eurydice/tree/protz/338 for now.

protz:
Great, thanks -- CI showed us that the bug was triggered before I pulled in the updated version of krml

PR #346, opened 2025-11-18
Adding the option [no-const] to disable the const pointers in C

Lin23299:
I've made 3 changes for the `no-const` option:
1. making each possible `const_of_xxx` function returns `false`.
2. Adding a `known_builtin_no_const` table to related both `SliceIndexShared` and `SliceIndexMut` with `Eurydice_slice_index_mut`
3. Always using `false` for builting fat pointers (`Dst_Ref`)

### DIFF
--- a/include/eurydice_glue.h
+++ b/include/eurydice_glue.h
@@ -136,10 +136,10 @@ typedef struct Eurydice_mut_borrow_slice_i16_s {
 #define Eurydice_array_eq(sz, a1, a2, t)                                       \
   (memcmp((a1)->data, (a2)->data, sz * sizeof(t)) == 0)
 // core::cmp::PartialEq<&0 (@Slice<U>)> for @Array<T, N>
-#define Eurydice_array_eq_slice_mut(sz, a1, s2, t, _)                          \
-  (memcmp((a1)->data, (s2)->ptr, sz * sizeof(t)) == 0)
 #define Eurydice_array_eq_slice_shared(sz, a1, s2, t, _)                       \
   (memcmp((a1)->data, (s2)->ptr, sz * sizeof(t)) == 0)
+#define Eurydice_array_eq_slice_mut(sz, a1, s2, t, _)                          \
+  Eurydice_array_eq_slice_shared(sz, a1, s2, t, _)
 
 // DEPRECATED -- should no longer be generated
 #define core_array_equality__core__cmp__PartialEq__Array_U__N___for__Array_T__N___eq( \
--- a/lib/AstOfLlbc.ml
+++ b/lib/AstOfLlbc.ml
@@ -214,70 +214,35 @@ module RustNames = struct
         }
     | _ -> failwith "impossible"
 
-  let known_builtins =
+  let known_builtins no_const =
+    let ( || ) no_const_variant default =
+      if no_const then
+        no_const_variant
+      else
+        default
+    in
     [
     (* slices *)
-    parse_pattern "SliceIndexShared<'_, @T>", Builtin.slice_index_shared;
+    parse_pattern "SliceIndexShared<'_, @T>", Builtin.(slice_index_mut || slice_index_shared);
     parse_pattern "SliceIndexMut<'_, @T>", Builtin.slice_index_mut;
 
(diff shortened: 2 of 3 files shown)


protz:
@Lin23299 can you apply this to your branch?

```
diff --git a/Makefile b/Makefile
index 052b73f..8819949 100644
--- a/Makefile
+++ b/Makefile
@@ -48,7 +48,7 @@ build: check-karamel check-charon
 CFLAGS         := -Wall -Werror -Wno-unused-variable $(CFLAGS) -I$(KRML_HOME)/include
 CXXFLAGS       := -std=c++17

-test: $(addprefix test-,$(TEST_DIRS)) custom-test-array custom-test-for testxx-result check-charon check-libcrux test-libcrux
+test: $(addprefix test-,$(TEST_DIRS)) custom-test-libcrux-no-const custom-test-array custom-test-for testxx-result check-charon check-libcrux test-libcrux

 clean-and-test:
        $(MAKE) clean-llbc
@@ -130,6 +130,11 @@ custom-test-for: test-for

 # libcrux tests

+custom-test-libcrux-no-const: test/libcrux.llbc
+       mkdir -p out/test-libcrux-noconst
+       $(EURYDICE) --config test/libcrux/c.yaml -funroll-loops 16 \
+         $< --keep-going --output out/test-libcrux-noconst --no-const
+
 test-libcrux: test/libcrux.llbc
        mkdir -p out/test-libcrux
        $(EURYDICE) --config test/libcrux/c.yaml -funroll-loops 16 \
```

it will make issues apparent -- almost everything type-checks, except for a function that relies on array_eq and array_eq_slice_shared, see lib/PreCleanup.ml, I think the signatures of those functions need to abandon shared slices

protz:
(I can't push to your branch because github has issues right now)

protz:
ok great! I will review your PR in more detail -- for now it looks like it just needs some CI tweaks

include/eurydice_glue.h
```diff
@@ -136,7 +136,9 @@ typedef struct Eurydice_mut_borrow_slice_i16_s {
 #define Eurydice_array_eq(sz, a1, a2, t)                                       \
   (memcmp((a1)->data, (a2)->data, sz * sizeof(t)) == 0)
 // core::cmp::PartialEq<&0 (@Slice<U>)> for @Array<T, N>
-#define Eurydice_array_eq_slice(sz, a1, s2, t, _)                              \
+#define Eurydice_array_eq_slice_mut(sz, a1, s2, t, _)                          \
+  (memcmp((a1)->data, (s2)->ptr, sz * sizeof(t)) == 0)
+#define Eurydice_array_eq_slice_shared(sz, a1, s2, t, _)                       \
```
protz: please `#define eq_slice_shared(x, y, z) eq_slice_mut(x, y, z)` to avoid copy-paste

lib/AstOfLlbc.ml
```diff
@@ -255,6 +255,47 @@ module RustNames = struct
   ]
   [@ocamlformat "disable"]
 
+  let known_builtins_no_const =
```
protz: this is too much copy-paste and I'm concerned that this is going to be hard to maintain

here's a sketch of what I would do:

```
let known_builtins no_const =
  let (||) default no_const_variant = if no_const then no_const_variant else default in
  [
    parse_pattern "SliceIndexShared...", Builtin.(slice_index_shared || slice_index_mut);
```

and then propagate the extra parameter wherever needed

protz (changes requested):
I made a couple suggestions, and I fixed some details in the Makefile. Other than that, I think your PR looks good and I'm happy to merge it.

For now, I think we shouldn't bother having C++ tests for the non-const version, I think making sure a big test like libcrux still passes the non-const version (in terms of passing Checker) is enough. Let me know if you disagree

include/eurydice_glue.h
Lin23299: Fixed

lib/AstOfLlbc.ml
Lin23299: Fixed

Lin23299:
@protz Thanks for the review!  I think we are good now?

protz (approved):

PR #361, opened 2025-11-27
Update charon

N1ark:
Partner to https://github.com/AeneasVerif/charon/pull/920


protz:
fatal: Not a valid commit name 9f447cddf5ceaf5065f12e4a5c233ec497452005

PR #363, opened 2025-12-04
Split headers: `libcrux-iot` additions

jschneider-bensch:
These are the glue additions that make all the extractions build using the upstream glue only.


Nadrieril:
Weird, but I think this will get fixed when new commits get pushed to the branch.

protz:
Just to recap. I'm waiting on my comment to be fixed: https://github.com/AeneasVerif/eurydice/pull/363#discussion_r2611040593 so that we don't merge something with undefined behavior

PR #379, opened 2026-02-14
Bump Charon

N1ark:
Companion PR to https://github.com/AeneasVerif/charon/pull/1023

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -26,11 +26,11 @@
         "rust-overlay": "rust-overlay"
       },
       "locked": {
-        "lastModified": 1771254791,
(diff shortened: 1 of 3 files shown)


N1ark:
really we ignore `PlaceMention`s but they should be kept, i think; they may e.g. contain an array access, in which case we'd get `T _ = x[5]`
@protz opinons / guidance on how to do it, given a place?

N1ark:
done :) thanks for the feedback! it's what i had done initially but the `EIgnore` didn't appear in the resulting C so I assumed it was maybe not needed, I re-added it

PR #388, opened 2026-04-07
Signed integer arithmetic in generated C is undefined behavior

nadimkobeissi:
In C, signed integer overflow is undefined behavior. In Rust, it is defined (panic or wrap). Eurydice translates Rust's signed `+`, `-`, `*` to plain C operators, which means `int32_t` and `int64_t` arithmetic in generated code is UB on overflow.

The standard fix is to cast to unsigned, operate, cast back:

```c
// Before (UB on overflow):
int32_t r = a + b;

// After (defined, identical machine code):
int32_t r = (int32_t)((uint32_t)a + (uint32_t)b);
```

This generates the same `add` instruction but gives the program defined semantics under the C standard.

### Proof of Concept

Compiling with `-fsanitize=signed-integer-overflow` confirms the UB:

```
runtime error: signed integer overflow: 2000000000 + 2000000000
    cannot be represented in type 'int32_t' (aka 'int')
```

And at `-O2`, the compiler exploits the UB. This overflow check:

```c
int32_t sum = a + b;
if (b > 0 && sum < a) return 1; // overflow
return 0;
```

returns `0` for `a = b = 2000000000` because the compiler reasons that `a + b >= a` when `b > 0` (since signed overflow "can't happen"). The fixed version using unsigned casts returns `1`.

### Note on `int16_t`

Due to C's integer promotion rules, `int16_t + int16_t` promotes both operands to `int` (32-bit) before the operation, so the arithmetic itself doesn't overflow — only the narrowing back to `int16_t` occurs, which is implementation-defined, not UB. The real UB is in `int32_t` and `int64_t`.

### Changed files

**`AstOfLlbc.ml`**: `mk_op_app` now wraps signed `Add`/`Sub`/`Mult`/`Neg` in unsigned casts, so generated C uses `(int32_t)((uint32_t)a + (uint32_t)b)` instead of `a + b`.

**`eurydice_glue.h`**: Adds `wrapping_add`/`wrapping_sub`/`wrapping_mul`/`wrapping_neg` for `i8`, `i16`, `i32`, `i64` (previously only unsigned types had these). Fixes `core_ops_arith__i32__add_assign` which had the same UB.

**`Cleanup1.ml`**: Extends unsigned wrapping inlining to `wrapping_sub` and `wrapping_mul` (previously only `wrapping_add`).

### Operational note

While I am confident in the existence of this bug, and while I have made every effort to guarantee that my implemented fix follows Eurydice engineering practices, I am not completely certain that I've implemented it in the way in which it is expected to implement such a change in Eurydice. This is due to Eurydice's relative complexity as a codebase and undocumented development practices. Review is strongly appreciated.


nadimkobeissi:
@protz I've reworked my contribution to be more palatable to you given your current priorities. Please review: https://github.com/AeneasVerif/eurydice/pull/388/changes/28f9fb069e81a008fa0fdc6b2a089331ccc16936

Although, if you'll permit me, I do have two points of contention with your position:

> I also think it'd be good to state in README.md the informal Eurydice guarantee

Do you really document such fundamental formal limitations only in the README? There should be a more formalized, unified place where such limitations of Eurydice are described in detail. These seem like important things to keep track of and to disclose clearly to all Eurydice users. 

> I would leave it as future work to implement a mode where Rust panics are transformed into C panics, not only because I'm not sure anyone wants this, 

I'm not a big Eurydice user, but I would assume that not only this is wanted by a general audience, but assumed to be true!

protz:
I'll respond in detail to your points tomorrow -- I'm running out of time for today.

But regarding your PR: can you make sure there is a test somewhere that exercises the functions that you are adding in eurydice_glue.h? Just want to make sure this is covered.

Thank you.

protz:
> Do you really document such fundamental formal limitations only in the README? There should be a more formalized, unified place where such limitations of Eurydice are described in detail. These seem like important things to keep track of and to disclose clearly to all Eurydice users.

I'm happy to create a tracking issue, and I think between your clarification to README and the existing "About" description on GitHub that says "Verify programs in Rust, still get C code for legacy environments.", we should be good.

> I'm not a big Eurydice user, but I would assume that not only this is wanted by a general audience, but assumed to be true!

Perhaps; we don't know. But what I mean is that I cannot realistically allocate a significant amount of my time for authoring or reviewing such changes, unless there is a motivating use-case.

PR #406, opened 2026-05-18
Two spots were missing correct const-ness for raw pointers

protz:


protz:
Found while trying to extract compiler-builtin's uint128 implementation.

PR #407, opened 2026-05-19
Update karamel with FStarLang/karamel#741

mtzguido:
Generated code has a few more casts. This will be improved in a follow up karamel patch.


protz:
Great. I think the improvements would be nice :-)

PR #410, opened 2026-05-28
Update charon

Nadrieril:
Companion to https://github.com/AeneasVerif/charon/pull/1192


protz:
(heads-up @jschneider-bensch)

PR #426, opened 2026-07-03
Order declarations following charon-provided order

Nadrieril:
@protz this uses the declaration order provided by Charon, which is much more stable than a toposort based on a tree traversal, as that can be affected by small changes in the control-flow. That's a massive test diff though, so I wanna check with you first.

AI disclaimer: written by an LLM with opinionated input from me


lib/DeclOrder.ml
```diff
@@ -0,0 +1,94 @@
...
+let record_file (_, decls) = record_decls decls
+
+let inherit_order ~from_lid ~to_lid =
+  match Hashtbl.find_opt order from_lid with
+  | Some order_id when not (Hashtbl.mem order to_lid) ->
+      Hashtbl.add order to_lid order_id;
+      true
+  | _ -> false
+
+let refresh_monomorphized_lids () =
```
protz: is this fixpoint calculation needed? can you do a single pass and see what breaks?

lib/DeclOrder.ml
```diff
@@ -0,0 +1,94 @@
...
+      Krml.MonomorphizationState.state;
+    Hashtbl.iter
+      (fun (generic_lid, _, _) monomorphized_lid ->
+        changed := inherit_order ~from_lid:generic_lid ~to_lid:monomorphized_lid || !changed)
+      Krml.MonomorphizationState.generated_lids
+  done
+
+let order_of_decl decl = Hashtbl.find_opt order (lid_of_decl decl)
+
+let compare_order current_pos1 decl1 current_pos2 decl2 =
```
protz: how do we end up in the case where we compare order for something that was not registered before? does that not indicate a bug?

protz (commented):
Overall I'm in favor of this but it seems like there are a lot of defensive cases that, intuitively, should not happen. Can you investigate a little bit which are needed and put asserts for whatever doesn't trigger?

lib/DeclOrder.ml
Nadrieril: Single pass does seem to work

lib/DeclOrder.ml
Nadrieril: Turns out karamel uses some virtual lids for equality helpers and these don't exist in Charon. We can register them early, it's clearer

Nadrieril:
Alright I made this into a real PR by rewriting most of it, now it's actually reviewable. The one bit I haven't looked into is the change to the topological sorting algorithm, I assumed the LLM knows how to implement that kind of stuff.

protz (commented):
Generally looks good, although I was unfamiliar with Kahn's algorithm.

Do I understand correctly that we first sort declarations using the Charon order, then do something smart (the "ranking") to give a meaningful order to the declarations at the end of the transformation pipeline?

Why are indegrees and successors hash tables of references? Can't they just be regular hash tables and you call `Hashtbl.replace` to mutate the value in place rather than having two levels of indirection?

@jschneider-bensch heads-up that this change is coming, please give it a try if you can to catch any breakage early... thanks

protz (commented):
More thoughts. Is it that Kahn's algorithm is a stable topological sort? If so, I think we should articulate why we choose this particular algorithm over the simpler traversal

Nadrieril:
Yeah the toposort I didn't look too much into. The reason we need it is that Karamel is still doing the monomorphization, and apparently the topological order given by charon on poly items is not sufficient to get a topological order on mono items.

> Do I understand correctly that we first sort declarations using the Charon order, then do something smart (the "ranking") to give a meaningful order to the declarations at the end of the transformation pipeline?

Not exactly: the ranking is not smart, it's just recording the Charon order on poly items. Then the smart part is managing to recover the order of a mono item, by remembering which poly item it came from. Once we've done that, then we need to add some toposorting on top and that's where the vibecoded stable toposort comes in.

> If so, I think we should articulate why we choose this particular algorithm over the simpler traversal

What's "the simpler traversal"?

Nadrieril:
Cleaned up the toposort, it's decent code now

protz (approved):

protz:
Thanks this is great! This should make life much easier for downstream users

PR #427, opened 2026-07-06
DRAFT: Use Charon's monomorphization mode

Nadrieril:
This is an LLM-generated migration to Charon's `--monomorphize`, which should give us an idea of how far we are to that. I left most of the poly handling code, to make the diff smaller.

The bulk of the diff is:
- Builtins, because we couldn't easily reuse the existing karamel-polymorphic ones;
- Name handling;
- Making the per-file splitting stuff still work with this new monomorphization approach.

I think one thing that would help the migration would be having karamel names be less string-based. For example, if karamel represented monomorphized names like Charon does (with full type info), we could have Charon generate the same names as today, if that matters.

cc @protz, curious for your opinion on this


protz (commented):
Some general comments (I did not look at this in detail).
- The long names are back, which is something that I was explicitly trying to avoid because it gives bad code quality for libcrux. Is there a way to replicate a comparable "short names" mechanism in Charon?
- Are there opportunities for removing code? If everything is monomorphized, do we still need const-generic monomorphization in the krml pipeline?
- Lots of changes intertwined so I'm having a hard time making sense of it all.
- I'm pretty positively impressed that this works! This is excellent and will solve many other issues that were dependent on Charon-side monomorphization

Nadrieril:
> * The long names are back, which is something that I was explicitly trying to avoid because it gives bad code quality for libcrux. Is there a way to replicate a comparable "short names" mechanism in Charon?

Charon has such a mechanism already but I don't know if it does the same. Can we reuse the same code maybe? Was it karamel doing that computation?

> * Are there opportunities for removing code? If everything is monomorphized, do we still need const-generic monomorphization in the krml pipeline?

There are a ton of opportunities for removing code. You should not need const-generic monomorphization in krml since Charon does all of that already.

> * Lots of changes intertwined so I'm having a hard time making sense of it all.
> * I'm pretty positively impressed that this works! This is excellent and will solve many other issues that were dependent on Charon-side monomorphization

Yeah sorry this was meant as illustration more than anything, I wouldn't want to propose this as a PR since I don't really understand any of it.

PR #428, opened 2026-07-06
Update ocamlformat to 0.29.0

mtzguido:
I cannot install 0.27.0 on a recent OCaml after an opam update+upgrade.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -168,7 +168,7 @@
               pkgs.gnumake
               pkgs.llvmPackages_18.clang-tools # For clang-format
               pkgs.ocamlPackages.ocaml
-              pkgs.ocamlPackages.ocamlformat_0_27_0
+              pkgs.ocamlPackages.ocamlformat_0_29_0
               pkgs.ocamlPackages.dune_3
             ];
           } ''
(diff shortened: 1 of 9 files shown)


flake.nix
```diff
@@ -168,7 +168,7 @@
               pkgs.gnumake
               pkgs.llvmPackages_18.clang-tools # For clang-format
               pkgs.ocamlPackages.ocaml
-              pkgs.ocamlPackages.ocamlformat_0_27_0
+              pkgs.ocamlPackages.ocamlformat_0_29_0
```
Nadrieril, line 171: This is avaialble but only on a recent enough nixpkgs. If you have nix locally, try `nix flake update nixpkgs`.

protz: @Nadrieril do you know why CI fail shere?

Nadrieril: As I wrote just above, `ocamlformat_0_29_0` needs a more recent nixpkgs version

Nadrieril:
I fixed the nix situation with https://github.com/AeneasVerif/eurydice/pull/433

mtzguido:
Thanks @Nadrieril ! I'm not too familiar with Nix so I was gonna do some research before a patch, but very happy to be preempted : )

